PR #295, opened 2026-04-07
fix: resolve upload duplicate-key errors with generation counter

Destrayon:
## What

Eliminates all duplicate-key (23505) and FK-violation (23503) errors during concurrent uploads, re-uploads, and deletes by adding a **generation counter** to documents — a pattern used by LlamaIndex, LangChain, and Azure AI Search.

## Why

The original approach tried to handle concurrent upload races via exception catching (stale-row fallback lookups, `DbUpdateException` recovery). This was fragile — each fix exposed a new race window. The generation counter makes the system **self-healing**: even if cancellation races or misses, stale jobs harmlessly self-terminate.

## How

- **`DocumentEntity.Generation`** — monotonically increasing integer (default 1), incremented atomically via `INSERT...ON CONFLICT DO UPDATE SET generation = documents.generation + 1`
- **`StoreAsync` returns `StoreResult(DocumentId, Generation)`** — callers get the winning row's ID and current generation
- **`IngestionJob` + `IngestionOptions` carry `Generation`** — `0` means "skip check" (reindex/watcher/sync), non-zero triggers stale detection
- **Two generation checks in `IngestionPipeline`:**
  1. At start — before parsing/chunking/embedding (skips immediately)
  2. Before chunk insertion — catches re-uploads during long embedding calls
- **Removed** fallback/recovery code (stale-row lookup by path, `DbUpdateException` catch, `IsDuplicatePathConflict`)
- **Progress interface promotion** — `UpdateJobStatus`, `GetAllStatuses`, `RegisterJobCancellation`, `UnregisterJobCancellation` moved from concrete `IngestionQueue` to `IIngestionQueue` interface

## Test plan

- [x] Upload a file — ingests successfully with `gen=1`
- [x] Delete mid-ingestion — worker logs cancellation, no FK violations
- [x] Re-upload while ingesting — old job skipped (generation mismatch), new job completes
- [x] Rapid re-uploads — no 23505 errors, only latest version ingested
- [x] All 584 unit tests pass

Closes #295

### DIFF
--- a/src/Connapse.Core/Interfaces/IDocumentStore.cs
+++ b/src/Connapse.Core/Interfaces/IDocumentStore.cs
@@ -2,7 +2,7 @@ namespace Connapse.Core.Interfaces;
 
 public interface IDocumentStore
 {
-    Task<string> StoreAsync(Document document, CancellationToken ct = default);
+    Task<StoreResult> StoreAsync(Document document, CancellationToken ct = default);
     Task<Document?> GetAsync(string documentId, CancellationToken ct = default);
     Task<IReadOnlyList<Document>> ListAsync(Guid containerId, string? pathPrefix = null, int skip = 0, int take = 50, CancellationToken ct = default);
     Task DeleteAsync(string documentId, CancellationToken ct = default);
--- a/src/Connapse.Core/Interfaces/IIngestionQueue.cs
+++ b/src/Connapse.Core/Interfaces/IIngestionQueue.cs
@@ -37,6 +37,31 @@ public interface IIngestionQueue
     /// Gets the current queue depth.
     /// </summary>
     int QueueDepth { get; }
+
+    /// <summary>
+    /// Updates the status of a job (phase, progress, error).
+    /// </summary>
+    void UpdateJobStatus(
+        string jobId,
+        IngestionJobState state,
+        IngestionPhase? currentPhase = null,
+        double percentComplete = 0,
+        string? errorMessage = null);
+
+    /// <summary>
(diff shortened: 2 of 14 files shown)


Destrayon:
### Code review

Found 1 issue:

1. **Chunk purge uses the wrong document ID on the stale-row fallback path.** When `FindAsync(documentId)` returns null and the fallback `FirstOrDefaultAsync` by `(container, path)` finds a stale row (lines 126-132), `documentEntity` is reassigned to the stale row (which has a different `Id`), but `documentId` — used for the chunk purge at line 199 — is never updated. The purge `Where(c => c.DocumentId == documentId)` targets the *new* upload's GUID, which has no chunks yet, so the purge is a no-op. Stale chunks from the old row survive, doubling the chunk count on each re-upload. The same issue applies in the `DbUpdateException` recovery path (line 189 sets `isReindex = true` but `documentId` still refers to the original GUID). The code comment on line 196-197 explicitly warns: "Without this, every re-index doubles the chunk count, polluting search results." Fix: use `documentEntity.Id` instead of `documentId` in the chunk purge query, or reassign `documentId = documentEntity.Id` after each fallback/recovery reassignment.

https://github.com/Destrayon/Connapse/blob/88e9114fc52df0fa2720d589506bd22bde8d7b6a/src/Connapse.Ingestion/Pipeline/IngestionPipeline.cs#L196-L201

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #297, opened 2026-04-08
refactor: derived state pattern for optimistic bulk delete

Destrayon:
## Summary

Replaces brittle flag-based coordination for delete operations with a derived state pattern using `_pendingDeleteIds`. Follows Figma's LiveGraph "shadow state" approach:

```
DisplayedList = ServerState - PendingDeletions
```

### What changed

- Added `HashSet<string> _pendingDeleteIds` — entries in this set are filtered from `VisibleEntries` (the rendered list)
- `BulkDeleteSelected()` adds all selected IDs to the set upfront (entries vanish immediately), then deletes server-side. On failure, removes from set (entry reappears)
- `DeleteEntry()` — same pattern for single file delete
- `RefreshEntries()` — preserved-entry logic also filters out pending deletes
- Render loop, `ToggleSelectAll`, `AllSelected`, empty-folder check all use `VisibleEntries`

### What was removed

- No `entries.Remove()` / `entries.Add()` during delete (no list mutation)
- No `Task.Yield()` render flushing hacks
- No `HandleIngestionProgress` guards/early-returns
- No timing-dependent boolean flag coordination

### Why

The previous approach (PR #296) had `entries` mutated by three concurrent actors (user delete, ingestion progress handler, RefreshEntries) coordinated by boolean flags. This caused Firefox-only failures and each fix created a new regression. The `_pendingDeleteIds` set is the single coordination mechanism — background handlers keep running normally but can't make pending-delete entries visible.

## Files changed

Only `src/Connapse.Web/Components/Pages/FileBrowser.razor`

## Test plan

- [x] Upload files, bulk delete while processing — files vanish immediately
- [x] No ghost entries after page refresh
- [x] Upload without deleting — files stay visible (no regression)
- [x] Works in Chrome and Firefox
- [x] Works through Cloudflare (www.connapse.com)

### DIFF
--- a/src/Connapse.Core/Interfaces/IDocumentStore.cs
+++ b/src/Connapse.Core/Interfaces/IDocumentStore.cs
@@ -2,7 +2,7 @@ namespace Connapse.Core.Interfaces;
 
 public interface IDocumentStore
 {
-    Task<string> StoreAsync(Document document, CancellationToken ct = default);
+    Task<StoreResult> StoreAsync(Document document, CancellationToken ct = default);
     Task<Document?> GetAsync(string documentId, CancellationToken ct = default);
     Task<IReadOnlyList<Document>> ListAsync(Guid containerId, string? pathPrefix = null, int skip = 0, int take = 50, CancellationToken ct = default);
     Task DeleteAsync(string documentId, CancellationToken ct = default);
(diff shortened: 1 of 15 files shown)


Destrayon:
Closing — will recreate from clean main branch

Issue #54, opened 2026-03-03
Add CLI Reference and MCP Integration pages to wiki


Destrayon:
## Description
Create wiki pages documenting the Connapse CLI tool and MCP (Model Context Protocol) integration. CLI Reference covers installation, available commands, and usage examples. MCP Integration covers Claude Desktop setup and available tools.

## Acceptance Criteria
- [ ] **CLI Reference** page with:
  - Installation methods (npm global, from source)
  - Command reference with syntax, options, and examples
  - Global flags and configuration
  - Self-update instructions
  - Common workflows (container management, document upload, search)
- [ ] **MCP Integration** page with:
  - What MCP is and why it matters
  - Claude Desktop configuration (JSON config snippet)
  - Available MCP tools and their descriptions
  - Usage examples and tips
  - Troubleshooting common connection issues
- [ ] Sidebar updated with links

## Implementation Notes
- CLI commands defined in `src/Connapse.Cli/` — review for complete command list
- MCP tools defined in `src/Connapse.Mcp/` — review for complete tool list
- README has a brief MCP section that can be expanded
- CLI reference may need to be written largely from scratch

## Size Estimate
S — ~120-150 lines across 2 wiki pages

## Related
Parent: #52

Issue #53, opened 2026-03-03
Add Getting Started and Deployment guides to wiki


Destrayon:
## Description
Create the Getting Started and Deployment Guide wiki pages. Getting Started is the primary onboarding path for new users. Deployment Guide covers production Docker Compose setup, environment configuration, PostgreSQL+pgvector requirements, and backup/restore procedures.

## Acceptance Criteria
- [ ] **Getting Started** page with:
  - Prerequisites (Docker, .NET 9, PostgreSQL+pgvector)
  - Docker Compose quickstart (copy-paste ready)
  - Creating first container and uploading a document
  - Running first search query
  - Links to CLI and MCP pages for next steps
- [ ] **Deployment Guide** page with:
  - Full Docker Compose configuration reference
  - Environment variables table with descriptions and defaults
  - PostgreSQL setup (pgvector extension, connection strings)
  - Volume mounts and persistent storage
  - Backup and restore procedures
  - Production hardening tips (HTTPS, secrets management)
- [ ] Sidebar updated with links to both pages

## Implementation Notes
- Distill from `docs/deployment.md` (27K, very thorough) — wiki version should be more concise and task-oriented
- Getting Started content partially exists in README.md quickstart section
- Target audience: sysadmin or developer deploying Connapse for the first time

## Size Estimate
M — ~200-250 lines across 2 wiki pages

## Related
Parent: #52

Issue #52, opened 2026-03-03
Set up GitHub Wiki with structure, Home page, and sidebar navigation


Destrayon:
## Description
Enable the GitHub Wiki on the Connapse repository and establish the foundational structure: Home page with project overview and navigation, sidebar with organized page links, and footer with repo links. The wiki becomes the primary user-facing documentation hub, while `/docs/` remains the developer/architecture reference.

## Acceptance Criteria
- [ ] GitHub Wiki enabled on the repository
- [ ] Home page with project overview, feature highlights, and quick navigation
- [ ] `_Sidebar.md` with organized sections (Getting Started, Configuration, Integrations, Reference)
- [ ] `_Footer.md` with links to repo, issues, discussions, and project board
- [ ] README.md updated to point users to wiki for documentation
- [ ] `/docs/` files retained as developer reference (architecture, plans, API internals)

## Implementation Notes
- GitHub Wiki is a separate git repo: `https://github.com/Destrayon/Connapse.wiki.git`
- Pages are Markdown files at root level — no subdirectory support
- `_Sidebar.md` and `_Footer.md` are special navigation files
- Consider adding a note to `/docs/` README clarifying the docs vs wiki split

## Size Estimate
S — ~80-100 lines across Home + Sidebar + Footer

## Related
Parent issue for all wiki sub-issues

PR #309, opened 2026-04-18
fix: bind JWT iss to AS URL for RFC 9068 compliance

Destrayon:
## Summary

Binds the JWT `iss` claim to `{scheme}://{host}` (the authorization-server issuer identifier advertised by `/.well-known/oauth-authorization-server`) instead of the literal `JwtSettings.Issuer` string. Strict MCP clients validate the token's `iss` against AS metadata per RFC 9068 §2.2 / RFC 8414 and silently discard mismatches, which was masking an otherwise-successful token exchange.

## What changed

- `JwtTokenService.GenerateAccessToken`, `GenerateTokenPairAsync`, `RefreshTokenAsync` all accept an optional `issuer` override. Falls back to `JwtSettings.Issuer` if not supplied so existing callers / tests are unaffected.
- `OAuthEndpoints` computes `issuer = {scheme}://{host}` at both authorization-code and refresh-token grants and threads it through the token service.
- `IdentityServiceExtensions` sets `ValidateIssuer = false`. Signature validation plus the existing canonical-URI audience check in `OnTokenValidated` are sufficient; a static-string issuer can't express multi-hostname / prefix-mounted deployments.

Symmetric with the RFC 8707 audience binding in #308 — same pattern: the submodule stays tenant-unaware by deriving identity-server identity from the current request.

## Test plan

- [x] `dotnet test` — all 25 Identity tests pass (existing tests mint with `settings.Issuer` default, still validates in `ValidateToken`)
- [ ] Verify token's `iss` matches AS metadata via Claude Code CLI against a deployed server
- [ ] Custom MCP connector from claude.ai web remains blocked by [claude.ai-side bug](https://github.com/anthropics/claude-code/issues/46140) — out of scope for this PR

## References

- [RFC 9068 §2.2 — JWT Profile for OAuth 2.0 Access Tokens, iss claim](https://www.rfc-editor.org/rfc/rfc9068#section-2.2)
- [RFC 8414 — OAuth 2.0 Authorization Server Metadata](https://www.rfc-editor.org/rfc/rfc8414)
- [MCP Authorization spec 2025-06-18](https://modelcontextprotocol.io/specification/2025-06-18/basic/authorization)

## Summary by CodeRabbit

* **Improvements**
  * JWT validation now logs richer diagnostic warnings when audience checks fail, including request path and token audience details.
  * Authentication now enforces audience matching in custom validation while allowing per-request issuer evaluation for clearer multi-authority behavior.
  * Token issuance and refresh support an optional issuer override so minted tokens can reflect the request-derived issuer for better cross-origin portability.

### DIFF
--- a/src/Connapse.Identity/IdentityServiceExtensions.cs
+++ b/src/Connapse.Identity/IdentityServiceExtensions.cs
@@ -2,6 +2,7 @@ using System.Security.Cryptography;
 using System.Text;
 using Connapse.Core;
 using Connapse.Core.Interfaces;
+using Connapse.Core.Utilities;
 using Connapse.Identity.Authentication;
 using Connapse.Identity.Authorization;
 using Connapse.Identity.Data;
@@ -253,9 +254,12 @@ public static class IdentityServiceExtensions
                             .CreateLogger("JwtAud.Validate");
                         diagLogger.LogWarning(
                             "JWT audience mismatch at {Scheme}://{Host}{PathBase}{Path}: token auds=[{Audiences}] static='{StaticAud}'",
-                            request.Scheme, request.Host.Value, request.PathBase, request.Path,
-                            string.Join(", ", audList),
-                            jwtSettings.Audience);
+                            request.Scheme,
+                            LogSanitizer.Sanitize(request.Host.Value),
+                            LogSanitizer.Sanitize(request.PathBase.Value ?? string.Empty),
+                            LogSanitizer.Sanitize(request.Path.Value ?? string.Empty),
+                            LogSanitizer.Sanitize(string.Join(", ", audList)),
+                            LogSanitizer.Sanitize(jwtSettings.Audience));
 
                         context.Fail("Audience does not match this server.");
                         return Task.CompletedTask;
--- a/src/Connapse.Identity/IdentityServiceExtensions.cs
+++ b/src/Connapse.Identity/IdentityServiceExtensions.cs
@@ -254,7 +254,7 @@ public static class IdentityServiceExtensions
                             .CreateLogger("JwtAud.Validate");
                         diagLogger.LogWarning(
                             "JWT audience mismatch at {Scheme}://{Host}{PathBase}{Path}: token auds=[{Audiences}] static='{StaticAud}'",
-                            request.Scheme,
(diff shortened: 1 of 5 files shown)


Destrayon:
## CodeRabbit autofix run

Reviewed all 6 unresolved threads. Applied 1 fix in commit `4008196f1bd4091a5286b424f5716bde4946d6d5`.

### ✅ Applied

**🟠 Major — Sanitize log values with `LogSanitizer`** (`src/Connapse.Identity/IdentityServiceExtensions.cs:254`)

Wrapped `request.Host`, `PathBase`, `Path`, the joined audience list, and `jwtSettings.Audience` with `LogSanitizer.Sanitize()` per the project convention in `CLAUDE.md`. This also resolves the 3 CodeQL `cs/log-forging` alerts (#52, #53, #54) on the same log statement.

### ❌ Declined (with reasoning)

**🔴 Critical — Add explicit issuer validation** (`IdentityServiceExtensions.cs:188`)

`ValidateIssuer = false` is a deliberate design choice — the inline comment at lines 181–187 explains it, and it's the central change of this PR. The audience check at lines 243–248 already validates that any RFC 8707 resource-bound audience matches the request's `scheme + authority`. Per RFC 9068 §2.2, the `iss` claim equals the AS metadata URL (`{scheme}://{host}`), which is the same URL the audience binds to. So the existing audience check transitively enforces the issuer invariant for the resource-bound flow, and the static-audience legacy flow is anchored by the signing-key check.

Adding `ValidIssuer = jwtSettings.Issuer` would re-introduce the exact multi-host / reverse-proxy-prefix problem this PR was opened to fix — a static issuer string can't express a deployment reachable via multiple hostnames or mounted under a tenant slug. CodeRabbit's suggested `OnTokenValidated` issuer check is also redundant with the audience scheme+authority check it sits next to.

**🟡 Minor — Clarify `ValidateToken` issuer policy** (`Services/JwtTokenService.cs:27`)

`ValidateToken` already has an inline comment block at lines 162–167 explaining that the helper is intentionally standalone (no HTTP context, so no RFC 8707 audience check) and that callers needing per-deployment audience checks must go through the auth pipeline. The same constraint applies to issuer validation by extension. This is a doc-clarity nit, not a bug — the design rationale is captured at the call site.

Issue #352, opened 2026-08-14
Phase 4 — Sources tab, Connections settings, and Home simplification


Destrayon:
Part of #348. Blocked by #351.

## What

Add the Sources tab and the Connections settings tab, simplify the container create flow, and add source scoping to search.

## Why

This is the phase that actually fixes the three problems in #348. Until the file browser stops serving external connectors, the ownership framing and the browse-level permission leak both persist.

## How

Ships as **three PRs**:

| PR | Content | Est. |
|---|---|---|
| 4a | Connections settings tab | ~300 |
| 4b | Sources tab | ~300 |
| 4c | `Home.razor` simplification + Search facet | ~200 |

- **New `Sources.razor`** listing name, connection, scope, last sync time, status, document count, and errors. Actions: sync now, edit scope, enable/disable, delete. **No file listing** — content is reachable only through search, which already enforces per-user cloud scope.
- **New Connections tab** in Settings for connection CRUD, reusing the existing `S3ConnectionTester` and `AzureBlobConnectionTester` unchanged.
- **`Home.razor` gets simpler.** The create-container modal stops asking which connector to use and drops its S3/Azure/Filesystem config fields and Test Connection button — creating a container now always means managed storage.
- `FileBrowser.razor` serves only managed containers, so some read-only branching can go.
- `Search.razor` gains an owner-kind badge on results and a source facet for scoping queries.

Accepted residual risk, agreed during design: an aggregate document count is still a weak signal about a source's contents. Judged an acceptable trade for the tab being useful at all.

## Done when

No external connector is reachable through the file browser.

Destrayon:
Carried from the PR 3a planning for #351.

**Remove `FilesystemConnectorConfig.AllowUpload` / `AllowDelete` / `AllowCreateFolder` as part of this phase's `FileBrowser.razor` rework.**

The epic scopes these flags for removal in Phase 3, but they turned out to be UI config rather than enforcement config. They are read in more than ten places in `FileBrowser.razor` — the create-folder and upload buttons, the empty-state copy, the bulk-delete bar, per-row delete affordances, and a settings panel with an editable checkbox for each. Removing them in a backend contract PR would have pulled a substantial UI rework along with it.

They are already inert. Phase 2's backfill moved every Filesystem container to a source, and PR 3a deletes `ContainerWriteGuard`, which was the only thing that enforced them. What remains is dead config deciding which buttons render for containers that can no longer be of that type.

[...]

Issue #351, opened 2026-08-14
Phase 3 — Connector contract split and cursor-based sync engine


Destrayon:
Part of #348. Blocked by #350.

## What

Split write capability out of `IConnector`, add cursor-based incremental sync, rebuild the watcher around sources, delete `ContainerWriteGuard`, and add the `/api/sources` endpoints plus the `kind` discriminator.

## Why

The current sync model polls every five minutes, lists the entire remote corpus, and diffs it against an in-memory snapshot. That is O(corpus) per poll, it discards the durable cursor nearly every provider API offers, and it does not survive a restart.

## How

Ships as **three PRs**:

| PR | Content | Est. |
|---|---|---|
| 3a | `IWritableConnector` split, delete `ContainerWriteGuard`, drop Filesystem write flags | ~250, mostly deletions |
| 3b | `ISyncCursorConnector` + `SourceSyncService` | ~300 |
| 3c | `/api/sources` endpoints + MCP `kind` discriminator | ~250 |

- `WriteFileAsync`/`DeleteFileAsync` move to `IWritableConnector : IConnector`, implemented only by `MinioConnector`. A source becomes *incapable* of being written to rather than being told not to at runtime.
- **Delete `ContainerWriteGuard`** (~120 lines plus `ContainerWriteGuardTests`). It exists only because both kinds share one endpoint set; once routing separates them, it has no job.
- Add `ISyncCursorConnector` with `GetChangesAsync(string? cursor, ct)` returning `SyncDelta(Upserted, DeletedPaths, NextCursor, RequiresFullResync)`. `RequiresFullResync` exists because Microsoft Graph returns `HTTP 410 Gone` and Dropbox returns a `409 reset` on stale cursors — ignoring that produces a corpus that silently drifts out of sync.
- `ConnectorWatcherService` becomes `SourceSyncService`, re-keyed to sources. Its existing list-and-diff logic is retained as the explicit fallback for connectors without cursor support. S3 and Azure Blob stay on that path; neither offers a real delta API.
- Remove the Filesystem `AllowUpload`/`AllowDelete`/`AllowCreateFolder` flags and their guard branches.

## Done when

Sync runs off source rows, cursors survive a process restart, and sources expose no write route at all.

Destrayon:
Two CodeRabbit findings from #355 were deferred to this phase, because both become reachable exactly when the sync engine starts writing source-owned data.

[...]

PR #356, opened 2026-08-14
feat(storage): connection and source stores with encrypted secrets

### LINKED ISSUE #349: Phase 1 — Add connections and sources schema and stores
Part of #348. Implementation plan: `docs/superpowers/plans/2026-08-13-connector-source-split-phase-1.md`

## What

Add the `connections` and `sources` tables, bridge `documents` to both owner kinds, and expose `IConnectionStore` / `ISourceStore`. **Nothing in the running application reads them when this is done.**

## Why

Everything in #348 depends on a stable store API. Landing the schema first, with no consumers, means the risky data move in Phase 2 happens against tables that already exist and are already tested.

## How

Ships as **two PRs** — the phase is ~900 lines, past the `size/XL` "must split" threshold:

| PR | Content | Est. |
|---|---|---|
| 1a | `ConnectionEntity`, `SourceEntity`, `KnowledgeDbContext` config, owner bridge, EF migration, `OwnerBridgeSchemaTests` | ~450 |
| 1b | `PostgresConnectionStore`, `PostgresSourceStore`, DI registration, integration tests | ~450 |

Neither splits below ~450 without separating code from the tests that prove it.

Two details that are load-bearing and easy to get wrong:

- **The generated EF migration must be hand-corrected.** EF emits a drop-and-recreate for the renamed `chunks`/`chunk_vectors` owner columns, which would destroy every vector in the database. Replace it with catalog-only `ALTER TABLE ... RENAME COLUMN`, and drop the owner FKs via an `information_schema` lookup rather than hardcoded constraint names — the names differ between EF conventions and hand-written migrations.
- **`documents` gets a stored generated column.** `owner_id GENERATED ALWAYS AS (COALESCE(container_id, source_id)) STORED`, plus a CHECK that exactly one owner is set. This is what keeps the search path from ever having to classify its owner.

`ConnectionProvider` values deliberately match `ConnectorType` so the Phase 2 backfill is a direct cast. `ManagedStorage` is absent from `ConnectionProvider` — it is Connapse's own backend, never an external system it authenticates to.

## Done when

Migrations apply cleanly against both a fresh database and one with existing containers, documents, and vectors; `dotnet test` passes in full; both stores resolve from DI; and `git diff main --stat -- src/Connapse.Web/` is empty.

Destrayon:
Closes #349. Second of two PRs for Phase 1. **Stacked on #355 — review that first; this PR's diff shows only the store work.**

## What

Adds `PostgresConnectionStore` and `PostgresSourceStore` behind the `IConnectionStore` / `ISourceStore` interfaces, and registers both in DI. Still nothing in the running application reads them.

## Why

Phase 2's backfill needs a stable store API to write through. Landing the stores separately from the schema keeps each PR reviewable and lets the schema merge on its own.

## How

**Secrets are encrypted at rest** via ASP.NET Core DataProtection under the `Connection.v1` purpose, matching the existing `CloudIdentity.v1` pattern in `CloudIdentityService`. The `Connection` read model has no secret property at all — only a `HasSecret` boolean — and a unit test asserts that, so the value cannot leak through an API response by accident.

On update, an omitted secret **leaves the stored one intact** rather than wiping it. Without that, a settings form that doesn't re-send the password field would silently destroy the credential.

`PostgresSourceStore.UpdateSyncStateAsync` persists the sync cursor so incremental sync survives a process restart — the current in-memory snapshot does not. A null cursor is treated as a meaningful clear, not "no change", because that is exactly how a `RequiresFullResync` response (Graph's `HTTP 410 Gone`, Dropbox's `409 reset`) tells us to discard a stale delta token.

Deleting a connection that still has sources throws rather than cascading, backed by `DeleteBehavior.Restrict` at the database level.

### One thing worth a reviewer's opinion

This adds `Microsoft.AspNetCore.DataProtection` to `Connapse.Storage`, which previously had no ASP.NET Core dependency. `Connapse.Identity` solves the same problem with a full `FrameworkReference` to `Microsoft.AspNetCore.App`; I used the targeted package to keep the data layer off the framework. If you'd rather encryption not live in Storage at all, the alternative is an interface in `Connapse.Core` implemented in `Connapse.Web` — happy to refactor.

## Testing

- **All 1029 tests pass, 0 failures** (757 unit + 272 integration).
- 12 new integration tests across the two stores, each written first and confirmed failing before implementation: secret round-trip, no-secret returns null, update preserves an omitted secret, duplicate names rejected, cursor persistence, null-cursor clear, unknown-connection rejection, and connection-scoped listing.

## Summary by CodeRabbit

* **New Features**
  * Added connection management with creation, updates, deletion, pagination, and connection-scoped source listings.
  * Added secure credential storage and retrieval, with secrets excluded from connection details.
  * Added source management, configurable settings, summaries, and persisted synchronization status.
  * Added safeguards to prevent stale synchronization results from overwriting newer progress.
  * Added clear errors when protected connection credentials cannot be accessed.

* **Bug Fixes**
  * Preserved existing credentials when updating a connection without providing a replacement.
  * Prevented deletion of connections still associated with sources.

* **Tests**
  * Added integration coverage for connection security, source management, synchronization, and concurrency scenarios.

### DIFF
--- a/src/Connapse.Core/Interfaces/ISourceStore.cs
+++ b/src/Connapse.Core/Interfaces/ISourceStore.cs
@@ -12,10 +12,29 @@ public interface ISourceStore
     Task<bool> ExistsAsync(Guid id, CancellationToken ct = default);
 
     /// <summary>
-    /// Persists the sync cursor and outcome after a sync cycle. Passing a null
-    /// cursor clears it, which is what a RequiresFullResync response demands.
+    /// Unconditionally persists the sync cursor and outcome. Passing a null cursor clears
+    /// it, which is what a RequiresFullResync response demands.
+    /// <para>
+    /// This overwrites whatever is stored, so it is only safe when the caller is
+    /// deliberately resetting progress (a full resync, or recording a failure). For normal
+    /// forward advancement use <see cref="TryAdvanceSyncStateAsync"/>, which cannot clobber
+    /// a concurrent sync's newer cursor.
+    /// </para>
     /// </summary>
     Task UpdateSyncStateAsync(Guid id, string? cursor, SyncStatus status, string? error, DateTime? syncedAt, CancellationToken ct = default);
+
+    /// <summary>
+    /// Advances the sync cursor only if the stored cursor still equals <paramref name="expectedCursor"/>,
+    /// as a single atomic statement. Returns false when another sync already moved it, in which case
+    /// nothing is written and the caller should discard its result rather than retry blindly.
+    /// <para>
+    /// Without this, two overlapping syncs racing to completion can regress progress: if B starts
+    /// after A but finishes first, A's unconditional write replaces B's newer cursor and the next
+    /// cycle resumes from stale progress — duplicate ingestion at best, and an invalid continuation
+    /// token for providers whose cursors are opaque.
+    /// </para>
+    /// </summary>
+    Task<bool> TryAdvanceSyncStateAsync(Guid id, string? expectedCursor, string? newCursor, SyncStatus status, string? error, DateTime? syncedAt, CancellationToken ct = default);
 
     Task<ContainerSettingsOverrides?> GetSettingsOverridesAsync(Guid id, CancellationToken ct = default);
     Task SaveSettingsOverridesAsync(Guid id, ContainerSettingsOverrides overrides, CancellationToken ct = default);
--- a/src/Connapse.Core/Models/ConnectionModels.cs
+++ b/src/Connapse.Core/Models/ConnectionModels.cs
@@ -29,3 +29,18 @@ public record UpdateConnectionRequest(
     string? Name = null,
     string? ConfigJson = null,
     string? Secret = null);
+
+/// <summary>
+/// Thrown when a stored connection secret exists but cannot be decrypted — in practice,
+/// because the DataProtection key ring that encrypted it is gone or unreachable (a replaced
+/// volume, or a replica that does not share the key directory). Distinct from "no secret
+/// stored", which returns null. Callers should surface a reconnect prompt rather than
+/// treating this as a transient fault; retrying will not help.
+/// </summary>
+public class ConnectionSecretUnavailableException(Guid connectionId, Exception inner)
+    : Exception($"The stored secret for connection '{connectionId}' could not be decrypted. " +
+                "The DataProtection key ring that encrypted it is unavailable, so the credential " +
+                "must be re-entered.", inner)
+{
+    public Guid ConnectionId { get; } = connectionId;
+}
--- a/src/Connapse.Storage/Connapse.Storage.csproj
+++ b/src/Connapse.Storage/Connapse.Storage.csproj
@@ -20,6 +20,7 @@
     <PackageReference Include="System.Security.Cryptography.Xml" Version="10.0.10">
       <PrivateAssets>all</PrivateAssets>
     </PackageReference>
+    <PackageReference Include="Microsoft.AspNetCore.DataProtection" Version="10.0.2" />
     <PackageReference Include="Microsoft.Extensions.Http" Version="10.0.2" />
     <PackageReference Include="Microsoft.Extensions.Options.ConfigurationExtensions" Version="10.0.2" />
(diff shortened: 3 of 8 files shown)


Destrayon:
## Codex adversarial review

Ran an adversarial review of this branch against `feature/349-connector-source-split-phase-1a`. Verdict was **needs-attention** with two `[high]` findings. Both were verified against the code; they land differently.

### 1. Data Protection key ring — refuted

> Stored secrets can become permanently undecryptable after deployment or scale-out. The storage registration adds Data Protection without establishing a durable, shared key-ring contract... A container replacement can lose those keys.

Not accurate for this codebase. `Program.cs` already configures persistence and an application name:

```csharp
builder.Services.AddDataProtection()
    .PersistKeysToFileSystem(dpKeysDir)   // /app/appdata/DataProtection-Keys
    .SetApplicationName("Connapse");
```

That directory maps to the named `appdata` Docker volume precisely so keys survive container replacement. Codex reviewed the diff in isolation, saw an unconfigured `IDataProtectionProvider` being injected, and inferred an unconfigured host.

**What does survive the refutation**, and is worth tracking separately:

- Multi-host scale-out without a shared key volume would still fail to decrypt across replicas.
- There is no recovery UX when `Unprotect` throws after key loss — `IConnectionStore.GetSecretAsync` documents that it throws `CryptographicException`, but nothing surfaces that to an admin.

Both are **pre-existing**: `CloudIdentityService` has encrypted cloud identity tokens under `CloudIdentity.v1` through the same key ring since v0.3.0. This PR adds a second consumer, not a new risk class. The recovery UX belongs with the Connections tab in #352.

### 2. Sync cursor lost-update — valid, deferred to #351

> `UpdateSyncStateAsync` performs an unconditional update by source ID, with no expected-cursor/version predicate. If sync B starts after sync A but A finishes last, A overwrites B's newer cursor.

The mechanism is real. Nothing calls the method outside tests today — the sync engine lands in #351 — and the right fix depends on that engine's shape: a single scheduler loop per source makes overlap impossible, whereas a pooled or retrying worker needs compare-and-swap or a lease. `ISourceStore` has no other consumers, so changing the signature in Phase 3 costs nothing.

Recorded on #351 rather than guessing the concurrency contract before the consumer exists.

### Assessment

Neither finding blocks this PR. One is refuted on the facts; the other is a Phase 3 design decision with no current caller.

Destrayon:
Both Codex findings are now addressed in code (d7c5324), including the one I had argued was refuted.

**Cursor lost-update — fixed.** Added `ISourceStore.TryAdvanceSyncStateAsync`, a compare-and-swap issued as a single atomic `UPDATE ... WHERE` that advances the cursor only when the stored value still matches what the caller read. It returns false on conflict so a late-finishing sync discards its result instead of regressing progress.

One subtlety worth calling out: the predicate is built explicitly rather than comparing against a nullable parameter. `s.SyncCursor == expectedCursor` with a null parameter translates to `sync_cursor = NULL`, which is never true in SQL, so a first-ever advance would have reported a phantom conflict on every source.

`UpdateSyncStateAsync` stays unconditional and is now documented as the deliberate reset path. That distinction is load-bearing: a `RequiresFullResync` response (Graph `410 Gone`, Dropbox `409 reset`) must be able to clear the cursor regardless of its current value, so gating it behind compare-and-swap would break the very recovery it exists for.

Four new tests, including the reverse-completion-order case Codex asked for: two syncs read the same starting cursor, B commits first, A's late write is rejected and B's cursor survives.

**Key ring — partially addressed.** The core claim stays refuted; `Program.cs` persists keys to the named `appdata` volume with `SetApplicationName("Connapse")`. But the residual point about unhandled `Unprotect` failures was fair, so `GetSecretAsync` now wraps `CryptographicException` in `ConnectionSecretUnavailableException`, carrying the connection ID and logging at error level. A raw crypto exception reads as a corrupt payload when the real cause is nearly always a lost or unshared key ring, and the Connections UI in #352 needs to distinguish "re-enter this credential" from a transient fault.

The multi-host scale-out gap is unchanged and remains a deployment concern shared with `CloudIdentity.v1`.

All 1033 tests pass (757 unit, 276 integration).

Issue #348, opened 2026-08-14
Split managed containers from external sources


Destrayon:
## What

Split the single `Container` concept into three: **Connection** (an admin-registered credential for an external provider), **Source** (a read-only scope inside a connection that Connapse ingests), and **Container** (Connapse's own managed storage — the only thing that stays browsable and mutable).

Design spec: `docs/superpowers/specs/2026-08-13-connector-source-split-design.md`

## Why

Today every container is presented identically, whether it is managed storage or a mirror of S3, Azure Blob, or a local filesystem. All of them land in one grid on `Home.razor` and open the same file browser. Three problems follow:

1. **It implies ownership Connapse does not have.** Rendering an S3 bucket as a browsable folder tree tells the user these are Connapse's files. They are someone else's, mirrored at best.
2. **It leaks permissions.** The file tree exposes every synced object to any Connapse user regardless of whether they could see it in the source system. Browsing is a far wider surface than search results, which already pass through `CloudScopeService`.
3. **It clutters the mental model.** Synced buckets compete with real managed containers in the same list.

Writes are *already* gated by `ContainerWriteGuard`. The gap is presentation, not mutation.

## How

Five phases, each independently shippable and each leaving `main` in a working state:

- [x] #349 — Phase 1: Schema and stores (2 PRs) — merged in #355 and #356
- [x] #350 — Phase 2: Backfill and compatibility read — merged in #359
- [x] #351 — Phase 3: Connector contract and sync engine (3 PRs)
- [x] #352 — Phase 4: Sources and Connections UI (3 PRs)
- [x] #353 — Phase 5: Remove compatibility shims (1 PR)

Phases are the unit of scheduling; PRs are the unit of review. Three phases exceed the 300-line norm and land as multiple PRs, using `Part of #<phase>` on intermediate PRs and `Closes #<phase>` on the last. Roughly ten PRs total, merged to `main` as they land rather than stacked on a long-lived epic branch.

### Key decisions from the spec

- **A separate `sources` table, not a `Kind` discriminator on `containers`.** A discriminator leaves sources carrying a folder tree they must never have, forces five permanently-NULL columns onto managed rows, and lets the schema express nonsense like `Kind=Managed, ConnectorType=S3`.
- **Filesystem becomes a source** and loses its `AllowUpload`/`AllowDelete`/`AllowCreateFolder` flags. The rule becomes uniform: if Connapse does not own the bytes, you do not mutate them through Connapse.
[...]

PR #359, opened 2026-08-15
feat(storage): backfill external containers into connections and sources

### LINKED ISSUE #350: Phase 2 — Backfill external containers into sources with compatibility read
Part of #348. Blocked by #349.

## What

Migrate every non-managed container into a connection plus source pair, repoint its documents to `source_id`, and delete its folder rows. Container endpoints keep serving them through a compatibility read for the duration of the window.

## Why

This is the only genuinely irreversible step in #348. It ships while the old read path still works, so a bad backfill is a rollback rather than an outage.

## How

Single PR, ~300 lines.

- One connection per distinct `(ConnectorType, credential set)`; one source per migrated container. `ConnectionProvider` values match `ConnectorType`, so the mapping is a direct cast.
- Transactional and resumable **per container** — a partial run must not leave a corrupt state.
- Folder rows for migrated containers are deleted; sources have no folder tree.
- Compatibility read: container endpoints resolve an id against `containers` first, then `sources`, so existing links and agent calls keep working.

## Done when

An install seeded with pre-existing S3, Azure Blob, and Filesystem containers returns **identical search results before and after** the backfill, verified by a dedicated integration test that seeds an old-shape database, runs the backfill, and asserts document counts, vector counts, and search parity. That test is what makes this phase safe to ship.

Destrayon:
Closes #350. Phase 2 of #348 — the one irreversible step in the epic.

## What

Migrates every non-managed container into a connection + source pair, repoints its documents to `source_id`, deletes its folder rows, and keeps pre-migration container IDs resolving through a compatibility read. Runs once at startup and is idempotent.

Plan: `docs/superpowers/plans/2026-08-14-connector-source-split-phase-2.md` (#358)

## Why

Phase 1 built the schema with no consumers. This is the step that actually moves data, so it ships while the old read path still works — a bad backfill is a rollback rather than an outage.

## How

**Migrated sources reuse the container's GUID.** This is the entire safety argument. `documents.owner_id` is a stored generated column, `COALESCE(container_id, source_id)`, so moving a document from one column to the other *under the same GUID* leaves `owner_id` byte-identical. Consequences:

- `chunks` and `chunk_vectors` — which denormalize that same owner — need **zero writes**, so no IVFFlat reindex on the largest table in the system.
- Search parity holds **by construction** rather than by careful copying.
- IDs held by agent prompts, CLI scripts, and bookmarks still resolve.

The alternative (fresh GUIDs plus an ID map) would have required rewriting owner columns across every chunk and vector row.

**Connections deduplicate by credential identity, not per container.** Ten buckets in one region under one role become one connection with ten sources. Dedup is by a deterministic connection name backed by the unique index on `connections.name`. Comparing the stored `ConfigJson` instead would silently fail — Postgres `jsonb` normalizes property order and whitespace, so it does not round-trip to what `JsonSerializer` emitted, and every container would get its own connection.

**No backfilled connection carries a secret.** S3 uses `DefaultAWSCredentials` or an assumed role, Azure uses managed identity, Filesystem needs none — so the migration never touches DataProtection.

**Failure handling.** Transactional per container with a fresh `DbContext` each, so one malformed legacy config cannot poison the batch; failures are collected and reported. The startup hook never blocks boot on failure, because the compatibility read means an un-migrated install still serves every request correctly.

### Deviation from the plan

The plan called for a `kind` discriminator on container responses. I did not add it: that changes the shape of *every* container response, which is an API contract change belonging with the MCP/REST work in #351. Instead the compatibility read projects a migrated source into the existing `Container` shape, so clients see an unchanged body. The fallback is also **reads only** — deleting or syncing a source through the containers route still 404s, which seems correct rather than accidental.

## Testing

**1053 tests pass, 0 failures** (766 unit, 287 integration). 20 new, all written test-first.

Automated coverage: config mapping and dedup-key stability; ID preservation; document repointing with `owner_id` unchanged; folder deletion; shared-credential dedup; managed containers untouched; idempotency on a second run; the compatibility read resolving a migrated ID and still 404ing an unknown one; and three parity tests asserting chunk, vector, and document ownership are unchanged and embeddings survive byte-identical.

**Manual verification against real pre-existing data**, since Testcontainers always starts empty and that is where a migration actually bites. Provisioned a pgvector instance, seeded five containers (two S3 sharing a region, one Azure, one Filesystem, one managed) plus documents, a chunk, an embedding, and a folder row, then ran the application against it:

```
containers remaining:  managed-keep only
sources created:       4, all with their original GUIDs
connections:           3 — s3-alpha and s3-beta correctly share s3-us-east-1
secrets:               none (no_secret = t on all three)
documents:             owner_id identical to the old container_id on every row
chunks/vectors:        owner untouched, embedding still [1,0,0]
folders:               0 remaining for the migrated container
```

### DIFF
--- a/src/Connapse.Core/Interfaces/IConnectorConfigMapper.cs
+++ b/src/Connapse.Core/Interfaces/IConnectorConfigMapper.cs
@@ -1,0 +1,13 @@
+namespace Connapse.Core.Interfaces;
+
+public interface IConnectorConfigMapper
+{
+    /// <summary>
+    /// Splits a legacy containers.connector_config blob into the connection identity
+    /// (credential + endpoint, shared across containers) and the source scope (the
+    /// specific bucket/prefix/subpath this container pointed at).
+    /// Throws ArgumentException for ManagedStorage, which is never migrated.
+    /// </summary>
+    (ConnectionIdentity Connection, string ScopeJson) Map(
+        ConnectorType type, string? connectorConfig, string containerName);
+}
--- a/src/Connapse.Core/Models/BackfillModels.cs
+++ b/src/Connapse.Core/Models/BackfillModels.cs
@@ -1,0 +1,27 @@
+namespace Connapse.Core;
+
+/// <summary>
+/// The credential-and-endpoint identity of a connection, extracted from a legacy
+/// container's connector config. DedupKey is what decides whether two migrated
+/// containers share one connection: it covers the credential and endpoint only,
+/// never the scope (bucket, blob container, subpath).
+/// </summary>
+public record ConnectionIdentity(
+    ConnectionProvider Provider,
+    string DedupKey,
+    string Name,
+    string? ConfigJson);
+
+public record BackfillPlanItem(
+    Guid ContainerId,
+    string ContainerName,
+    ConnectorType ConnectorType,
+    ConnectionIdentity Connection,
+    string ScopeJson);
+
+public record BackfillReport(
+    int ContainersMigrated,
+    int ConnectionsCreated,
+    int DocumentsRepointed,
+    int FoldersDeleted,
+    IReadOnlyList<string> Failures);
--- a/src/Connapse.Storage/Backfill/ConnectorConfigMapper.cs
+++ b/src/Connapse.Storage/Backfill/ConnectorConfigMapper.cs
@@ -84,6 +84,15 @@ public class ConnectorConfigMapper : IConnectorConfigMapper
     /// config JSON instead — Postgres jsonb normalizes property order and whitespace on
     /// storage, so a round-tripped blob will not match what JsonSerializer produced and
     /// every container would end up with its own connection.
+    /// <para>
+    /// The readable slug alone is NOT safe to key on, because slugification is lossy in
+    /// two ways: every non-alphanumeric character collapses to '-' (so the distinct roles
+    /// ".../role/a" and ".../role-a" both become "role-a"), and the result is truncated to
+    /// varchar(128) (so two long ARNs differing only near the end would match). Either
+    /// collision would silently attach a source to another credential's connection. The
+    /// hash suffix makes the name injective with respect to the dedup key, and is appended
+    /// after truncation so it always survives.
+    /// </para>
     /// </summary>
     private static string NameFor(string provider, string dedupKey, string containerName)
     {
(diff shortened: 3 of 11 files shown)


Destrayon:
## Codex adversarial review

Ran an adversarial review against `main` focused on the irreversible parts: GUID reuse, transactional correctness, dedup, replica concurrency, and the compatibility read. Verdict was **needs-attention** with three `[high]` findings. Two were real and are fixed in a1a43b6; one was a misread.

### 1. Name-only dedup can attach a source to the wrong credentials — **valid, fixed**

The stated mechanism was slightly off — the connection name *does* encode the credential fields, since it is derived from the dedup key (`s3|{region}|{roleArn}`), so "two containers targeting the same named endpoint with different roles" already produced different keys. But the underlying concern was correct, because **slugification is lossy in two ways**:

- Every non-alphanumeric character collapses to `-`, so the genuinely distinct roles `arn:aws:iam::1:role/a` and `arn:aws:iam::1:role-a` both became `s3-us-east-1-arn-aws-iam-1-role-a`.
- The result is truncated to `varchar(128)`, so two long ARNs differing only near the end collided.

Either collision would silently bind the second source to the first connection and inherit the wrong credentials — exactly the failure described. Fixed by appending an 8-hex-char SHA-256 discriminator of the dedup key, applied *after* truncation so it always survives. Two regression tests cover both collision shapes.

### 2. Two replicas can race the irreversible backfill — **valid, fixed**

Correct. Every replica ran this at startup with no coordination, and the read-then-create path for connections is not concurrency-safe: two replicas migrating *different* containers that map to the same connection can both observe none and race. `RunAsync` now takes a `pg_try_advisory_lock` for the whole batch. It is non-blocking by design, so a losing replica skips entirely rather than partially migrating, and the winner does the work. New integration test holds the lock on a separate connection and asserts the second caller migrates nothing and leaves the container untouched.

### 3. Rollback does not reset the reused DbContext — **refuted**

This one misread the scope. The `DbContext` is not reused across containers:

```csharp
foreach (var row in legacy)
{
    try
    {
        ...
        // Fresh context per container so one failure cannot poison the next.
        await using var perItem = await factory.CreateDbContextAsync(ct);
        await using var tx = await perItem.Database.BeginTransactionAsync(ct);
```

`perItem` is created inside the loop body and disposed each iteration, so there is no surviving change-tracker state for a rollback to corrupt. The outer context is used only for the initial `AsNoTracking` read.

The review did confirm the thing actually worth confirming here: `ExecuteUpdateAsync` and `ExecuteDeleteAsync` do participate in the ambient `BeginTransactionAsync` scope when issued from the same `DbContext`.

## Testing

All **1056 tests pass, 0 failures** (769 unit, 288 integration) — 3 new for these fixes.

PR #363, opened 2026-08-15
feat(ingestion): make document ownership explicit and owner-aware

Destrayon:
Part of #351. First of two PRs for phase 3b. Plan: #362.

## What

Makes ingestion owner-aware, so a document can belong to a source rather than only to a container. Introduces `OwnerRef`, teaches `IngestionPipeline` to write `source_id`, and makes vector upserts demand an explicit owner.

Split out from the sync engine deliberately: these are two standalone bug fixes, and the 700-line `ConnectorWatcherService` rewrite deserves its own review rather than riding behind them.

## Why

Two bugs found while grounding the #362 plan, both of which would bite the moment sources are enumerated:

**1. Ingesting into a source was impossible.** `IngestionPipeline` set `documentEntity.ContainerId` unconditionally and never `SourceId`, so any source-owned document would have violated the `ck_documents_single_owner` CHECK constraint.

**2. Vectors could be written with no owner, silently.** `PgVectorStore` read an optional `containerId` from metadata and defaulted a missing or unparseable value to `Guid.Empty`:

```csharp
Guid containerId = Guid.Empty;
if (metadata.TryGetValue("containerId", out var containerIdStr))
    Guid.TryParse(containerIdStr, out containerId);
```

That is the worse of the two. The write succeeds, but every search path filters on `owner_id`, so a zero-owner vector matches nothing — the content is unreachable rather than visibly broken, and nothing surfaces an error.

## How

`OwnerRef(Guid Id, bool IsSource)` with `ForContainer` / `ForSource` factories, exposing `ContainerId` and `SourceId` that map exactly onto the two columns behind the CHECK constraint. Threading a bare `Guid` plus a boolean would have invited getting the pairing wrong at one call site out of several; one type makes it impossible.

`IngestionOptions` gains `Owner`, preferred over the existing `ContainerId` string, which stays so unrelated callers keep working.

Ownership is resolved **once, above the try block**, and every downstream write goes through it — the document row on both the create and reindex paths, the chunk's `OwnerId`, and the vector metadata key (now `ownerId`, not `containerId`). Above the try deliberately: the pipeline otherwise catches exceptions and records a "Failed" document, so validating inside would half-write a row for what is a programming error.

`PgVectorStore` now throws when `ownerId` is missing or unparseable, in both `UpsertAsync` and `UpsertBatchAsync`. Failing loudly at write time beats discovering it at search time.

## Testing

**1048 tests pass, 0 failures.** New coverage: `OwnerRef`'s mutual exclusivity, a source-owned document writing `source_id` with `container_id` null and `owner_id` matching, chunks carrying the source owner with no `Guid.Empty` present, and an ownerless call throwing rather than silently succeeding.

Eight existing `IngestionPipelineTests` broke because they relied on the old `Guid.Empty` fallback — they test pipeline mechanics like non-seekable streams and document-ID generation, not ownership, so they were **updated** with an explicit container owner rather than deleted. Unlike PR 3a, none of these scenarios became unreachable.

## Still to come in phase 3b

`SourceSyncService` replacing `ConnectorWatcherService` — the fix for external sources not syncing at all since #359. This PR is a prerequisite: the sync engine ingests with `OwnerRef.ForSource(...)`, which does not work until this lands.

## Summary by CodeRabbit

- **New Features**
  - Ingestion now supports ownership by either a container or a source.
  - Ownership information is preserved on ingested documents, chunks, and vector metadata.
  - Added sync results capturing updates, deletions, delta processing, resynchronization, and errors.

- **Bug Fixes**
  - Prevented ingestion without a valid owner.
  - Ensured source-owned content no longer falls back to an empty or incorrect container identifier.

### DIFF
--- a/src/Connapse.Core/Models/IngestionModels.cs
+++ b/src/Connapse.Core/Models/IngestionModels.cs
@@ -8,7 +8,15 @@ public record IngestionOptions(
     string? Path = null,
     ChunkingStrategy Strategy = ChunkingStrategy.Semantic,
     Dictionary<string, string>? Metadata = null,
-    int Generation = 0);
+    int Generation = 0)
+{
+    /// <summary>
+    /// Who will own the ingested document. Preferred over <see cref="ContainerId"/>, which
+    /// cannot express source ownership. When null, ContainerId is used and the owner is
+    /// taken to be a container.
+    /// </summary>
+    public OwnerRef? Owner { get; init; }
+}
 
 public record IngestionResult(
     string DocumentId,
--- a/src/Connapse.Core/Models/SyncModels.cs
+++ b/src/Connapse.Core/Models/SyncModels.cs
@@ -27,3 +27,22 @@ public record SourceSyncResult(
     bool UsedDeltaPath,
     bool RequiredResync,
     string? Error);
+
+/// <summary>
+/// Thrown when a reindex would move a document between ownership domains — source to
+/// container or the reverse. Distinct from ordinary ingestion failures because it must
+/// escape the catch-all that records a "Failed" document: a refused reindex is not the
+/// same as one that went wrong, and conflating them would leave the caller believing the
+/// work failed rather than that it was rejected.
+/// </summary>
+public class DocumentOwnershipChangedException(Guid documentId, OwnerRef existing, OwnerRef attempted)
+    : InvalidOperationException(
+        $"Document {documentId} is owned by {(existing.IsSource ? "source" : "container")} {existing.Id}; " +
+        $"refusing to reindex it as {(attempted.IsSource ? "source" : "container")} {attempted.Id}. " +
+        "Document ownership cannot change.")
+{
+    public Guid DocumentId { get; } = documentId;
+    public OwnerRef ExistingOwner { get; } = existing;
+    public OwnerRef AttemptedOwner { get; } = attempted;
+}
+
--- a/src/Connapse.Core/Models/SyncModels.cs
+++ b/src/Connapse.Core/Models/SyncModels.cs
@@ -1,0 +1,29 @@
+namespace Connapse.Core;
+
+/// <summary>
+/// Identifies who owns a document: a managed container or an external source. Exists so the
+/// ownership decision is made once, at the point the owner is known, rather than by threading
+/// a Guid and a boolean through every layer and hoping each call site pairs them correctly.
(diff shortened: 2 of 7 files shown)


Destrayon:
## Codex adversarial review

Verdict was **needs-attention** with one `[high]` finding. It was correct, it reproduced exactly as described, and verifying it exposed a second bug. Both fixed in 5fd9372.

### Reindex could silently transfer ownership — valid, fixed

> Reindexing a source-owned document with a container Owner or legacy ContainerId therefore clears SourceId and assigns ContainerId. Because exactly one owner remains populated, ck_documents_single_owner still passes, hiding the ownership transfer.

Exactly right, and the reason it is nasty is the second sentence: the CHECK constraint is satisfied either way, so the database cannot catch it. A source-owned document could be re-ingested under a container owner and cross an authorization boundary with nothing anywhere reporting a problem.

Reproduced with a test before fixing: ingest under a source, re-ingest the same `DocumentId` under a container owner, observe `source_id` cleared and `container_id` set.

Ownership is now immutable. Reindex loads the persisted owner and throws `DocumentOwnershipChangedException` on any mismatch, leaving both columns untouched.

**The exception type is load-bearing, not decoration.** The pipeline wraps everything in a catch-all that records a "Failed" document and returns zero chunks. A refused reindex is not a failed one — swallowing it would tell the caller the work went wrong rather than that it was rejected, and would leave the document's real owner looking broken. The catch-all rethrows this type specifically.

Getting that right took two attempts. Hoisting the check above the `try` also worked, but it moved the first `DbContext` touch outside the catch, which broke eight `IngestionPipelineTests` that use the InMemory provider and depend on its model-validation failure being caught. The dedicated exception keeps infrastructure failures catchable while letting ownership violations through.

### Second bug, found while verifying — fixed

`IngestByIdAsync` did `Guid.Parse(doc.ContainerId)`. `Document.ContainerId` is a string that is empty for source-owned documents, so reindex-by-id would have thrown a bare `FormatException` with nothing indicating why. It now fails with an actionable message noting that source documents are re-ingested through the sync engine.

This would have bitten immediately once PR 3b's sync engine started re-ingesting changed remote files.

## Testing

**1050 tests pass, 0 failures.** Two new: a rejected cross-domain reindex leaves both ownership columns unchanged, and a same-owner reindex — the normal re-sync case — still succeeds.

PR #364, opened 2026-08-16
feat(web): replace ConnectorWatcherService with SourceSyncService

### LINKED ISSUE #351: Phase 3 — Connector contract split and cursor-based sync engine

Destrayon:
Closes #351.

Sources have not synced since #359 merged. `ConnectorWatcherService` enumerated `containers` filtered to Filesystem/S3/AzureBlob, and the #350 backfill moved every one of those rows into `sources` — so it matched nothing and external syncing stopped altogether. This replaces it with `SourceSyncService`, keyed on sources.

## What changed

**`SourceSyncService`** — connectors implementing `ISyncCursorConnector` take a delta path; the rest fall back to listing the remote and diffing it against the indexed documents. Two improvements over the watcher: the fallback diffs against the **database** rather than an in-memory snapshot, so progress survives a restart; and a `RequiresFullResync` response deliberately takes the unconditional update path rather than compare-and-swap, because the provider has declared the stored token invalid and gating that clear would break the recovery it exists for. Forward progress uses compare-and-swap so a late-finishing cycle cannot overwrite newer progress.

**`ConnectorWatcherService` deleted** (700 lines), along with `CloudSyncTests`. Its four call sites were already no-ops — containers are managed storage only, and both watch methods return immediately for that type.

**Composite FK tying a chunk's owner to its document's.** `chunks.owner_id` and `chunk_vectors.owner_id` are plain denormalized copies; `documents.owner_id` is generated and cannot lie. Search filters on the copies, so a chunk with the wrong owner is returned to whoever owns that id. Divergence is not reachable through application code — one production path writes each column, both from the same `OwnerRef` that writes the document — but it is enforced anyway, because the consequence is a content leak and the guard is nearly free. Added `NOT VALID`, so every insert and update is enforced while existing rows are not scanned; a legacy diverged row cannot abort startup for operators whose data is fine.

## Bugs found while writing the tests

Tests were written before the implementation, and four of them failed against the first cut:

- **Five `PostgresDocumentStore` queries still filtered on `ContainerId`**, so every one returned nothing for a source-owned document. The visible symptom was deletion during sync silently doing nothing; the same blind spot hit anything looking up a source's documents by path.
- **The failure path rolled progress backwards** — it wrote back the cursor from the cycle-start snapshot, so a transient remote outage could wipe newer progress and force a full re-list. It now re-reads the stored value.
- **Change detection was missing.** `CloudSyncTests` covered skipping unchanged and in-flight files; the first cut enqueued every remote file every cycle. The content hash is computed *after* the download and parse, so it dedupes nothing that costs money — each source would have re-downloaded, re-chunked and re-embedded its entire remote every five minutes. The remote's size and timestamp are now stored on the document, which unlike the watcher's in-memory snapshot survives a restart.
- **`MapToModel` handed out an empty `ContainerId` for source-owned rows.** `container_id` is NULL there and `Nullable<Guid>.ToString()` returns `""` rather than throwing, so consumers silently got an empty owner and `StoreAsync`'s `Guid.Parse` of it would have thrown. Now maps `OwnerId`.

## Noted, not fixed

- `IDocumentStore.StoreAsync` always writes `container_id` and cannot express a source-owned row. No live caller needs it — the pipeline writes entities directly — but anything that later tries to pre-register a source document through it breaks silently.
- `FileBrowserChangeNotifier` now has no publisher. Its only one was the watcher, and it only ever fired for connector types that no longer exist as containers. #352 decides whether it gets one or is removed.

## Testing

`dotnet test` — **1058 passed, 0 failed.** Nine integration tests cover the sync service (both paths, cursor advance, compare-and-swap rejection, unconditional resync, failure handling, disabled sources, and change detection in both directions); two more cover the FK rejecting a diverged chunk and accepting a matching one.

Not yet run against a real remote — the plan calls for a LocalStack pass with a real S3 source before merge.

## Summary by CodeRabbit

- **New Features**
  - Added automatic synchronization for enabled sources, including change detection, deletion handling, cursor-based updates, and recovery from temporary remote errors.
  - Added read-only connector creation from source and connection settings across supported storage providers.
- **Bug Fixes**
  - Improved handling of source-owned documents and preserved remote metadata during reindexing.
  - Prevented chunks from being associated with the wrong document owner.
  - Added validation for connector settings, filesystem path boundaries, and AWS session-name limits.
- **Behavior Changes**
  - Replaced container-specific watchers with centralized source synchronization and reconciliation.


Destrayon:
Follow-up: the manual LocalStack check is now an automated test instead — `SourceSyncS3IntegrationTests`, three cases against a real S3 API (all objects enqueue, a deleted object removes its document, an untouched bucket is skipped on the second cycle).

These resolve the real `IConnectorFactory` from DI rather than substituting it, so they are the only coverage of a connection's credentials and a source's scope being recombined into a working connector — every other sync test fakes that step.

`dotnet test` — **1061 passed, 0 failed.** No manual pre-merge step remains.

Destrayon:
Two more gaps closed, plus one that turned out to be closed the wrong way round.

**`SyncAllAsync` had no coverage at all** — every other sync test entered at `SyncSourceAsync`, one level below the method the background loop actually calls, so the enumeration and the enabled filter were untested.

**The Azure equivalent of the LocalStack test isn't achievable.** Azurite can't authenticate `DefaultAzureCredential`, and `AzureBlobConnector` hardcodes `https://{account}.blob.core.windows.net`. Redirecting it would mean supporting shared-key auth — the stored cloud credential this project doesn't accept — so the gap stays open by design. Closed the reachable part instead: connector configs are now internal-readable and `SourceConnectorFactoryTests` asserts **every recombined field** for all three providers rather than just the resulting type. A dropped region or mis-joined prefix produces a connector of the right type and then reads the wrong bucket. The Azure `managedIdentityClientId` case carries the most weight — dropping it silently falls back to the default identity, which may have wider access than the source is meant to have.

**Still uncovered, worth knowing before merge:** the migration has only ever been applied to an empty `chunks` table, so the upgrade path on populated tables rests on `NOT VALID` skipping the scan rather than on a test. The delta path has no production implementer yet, so fakes are the only thing to test it against.

`dotnet test` — **1062 passed, 0 failed.**

Destrayon:
Found a way to test the migration on a populated database — `ChunkOwnerMigrationTests`.

EF can migrate to a *specific* migration, so the test owns its own PostgreSQL container (the shared fixture arrives already at head) and stages the real upgrade: migrate to the release before the constraint, write documents, chunks and vectors the way an existing deployment holds them — including a deliberately diverged chunk — then migrate the rest of the way.

This proves both halves of the `NOT VALID` decision:
- legacy divergence neither blocks the upgrade nor gets silently repaired
- a diverged row inserted *after* the migration is rejected

Non-vacuous by construction: seeding the diverged row would itself have thrown had the staged migration not actually stopped short of the constraint.

That closes the last gap I'd flagged as a merge risk. The only remaining untested path is the delta sync path, which has no production implementer yet — `ISyncCursorConnector` is a contract added in 3a that nothing implements, so fakes are the only thing available to test it against.

`dotnet test` — **1064 passed, 0 failed.**

PR #366, opened 2026-08-17
fix(storage): resolve links before confining a filesystem source to its root

### LINKED ISSUE #365: Filesystem source confinement is bypassable by a symlink or junction
## Summary

`ConnectorFactory.CombineUnderRoot` confines a source's `subPath` beneath its connection's `allowedRoot` using `Path.GetFullPath` plus a `StartsWith` prefix check. `Path.GetFullPath` is purely lexical — it never touches the filesystem — so it does not resolve symlinks or NTFS junctions. A reparse point placed inside an allowed root passes the guard, and `FilesystemConnector` then follows it during enumeration and ingests whatever is on the other side.

Anything the service account can read becomes ingestible, chunked, embedded, and retrievable through search and the MCP server — including `appsettings.json`, the DataProtection key ring that decrypts stored connection secrets, and SSH keys.

## Verified

Empirically, on Windows, .NET 10.0.303, against the exact APIs in use:

| Step | Result |
|---|---|
| `mklink /J root\escape secret\` (no elevation required) | Junction created |
| `Path.GetFullPath(root\escape)` | Not resolved — stays lexically under root |
| `combined.StartsWith(rootFull + separator)` | **True — guard passes** |
| `Directory.EnumerateFiles(root, "*", AllDirectories)` | **Returns `root\escape\keyring.txt`** — content from outside the root |

Both halves fail: the guard admits the path and the walk reads through it.

## Affected

- `src/Connapse.Storage/Connectors/ConnectorFactory.cs` — `CombineUnderRoot`, lexical-only containment check
- `src/Connapse.Storage/Connectors/FilesystemConnector.cs:79` — bare `SearchOption.AllDirectories` overload, no `EnumerationOptions`, so reparse points are followed

## Exposure today

Not remotely reachable: there are no connection or source write endpoints, so an attacker cannot currently create a filesystem connection. Reaching this requires the ability to write a symlink inside an already-configured root, or database access. It must be fixed before any source/connection write surface ships (phase 3c of #351).

## Fix

1. Resolve link targets before the containment check — `FileSystemInfo.ResolveLinkTarget(returnFinalTarget: true)`, then re-verify containment. Java's `getCanonicalPath` resolves symlinks; .NET requires this explicitly. PHP's `open_basedir` resolves symlinks before its check for exactly this reason.
2. Refuse reparse points during enumeration, and re-resolve per entry rather than trusting the one-time config-time check — the walk runs every five minutes for the life of the process, so a directory swapped for a symlink after validation wins (CWE-367 TOCTOU).
3. Constrain `allowedRoot` itself to a configured allowlist, following Elasticsearch's static `path.repo`. The allowlist stops `allowedRoot: "/"`; it does not stop a symlink — both controls are needed.

Bind mounts and hardlinks remain unfixable by canonicalization, since the resolved path genuinely is beneath the root. Those need OS-level controls: a dedicated low-privilege service account and not co-locating the app's own config or key ring anywhere reachable. Worth documenting as a deployment note.

## Background

Prior art and the full threat analysis: `docs/research/programmatic-source-configuration-safety-2026-08-17.md`. Related: CWE-22, CWE-59, CWE-367; CVE-2021-43798 (Grafana) is the same class introduced by a refactor.

Destrayon:
Closes #365.

Filesystem source confinement was bypassable by a symlink or NTFS junction. `Path.GetFullPath` is purely lexical — it never touches the filesystem — so it does not resolve reparse points. A junction inside an allowed root passed the guard, and `Directory.EnumerateFiles(.., SearchOption.AllDirectories)` then descended through it.

Impact: anything the service account could read became ingestible, chunked, embedded, and retrievable through search and the MCP server — including `appsettings.json` and the DataProtection key ring that decrypts stored connection secrets.

Verified on Windows against .NET 10.0.303 before the fix, against the exact APIs in use:

| Step | Result |
|---|---|
| `mklink /J root\escape secret\` (**no elevation required**) | Junction created |
| `Path.GetFullPath(root\escape)` | Not resolved — stays lexically under root |
| `combined.StartsWith(rootFull + separator)` | **True — guard passes** |
| `Directory.EnumerateFiles(root, "*", AllDirectories)` | **Returns `root\escape\keyring.txt`** |

Both halves failed: the guard admitted the path and the walk read through it.

## Two further holes found while fixing it

Both in the same method, and the first is arguably worse than the reported bug since it needs no filesystem setup at all:

- **`GetFullPath` returned any rooted path unchecked**, with the comment "for watcher events". Any caller passing an absolute path read straight past the root with no containment check. Watcher events originate inside the root and still resolve; ones that do not are exactly what this must refuse.
- **The prefix comparison had no trailing separator**, so `/data-other` passed as being inside `/data` — the nginx `alias` off-by-slash bug.

## The fix

New `PathConfinement` helper in `Connapse.Core.Utilities`, used by both `ConnectorFactory.CombineUnderRoot` and `FilesystemConnector.GetFullPath`:

- Resolves links on **every path segment, not just the leaf**. This is the subtle half: in `/root/link/keyring.txt` the file itself is not a link, so asking it for a link target returns nothing and the path looks contained — only the parent gives the escape away. My first attempt resolved only the leaf and the tests caught it.
- Compares with a trailing separator, and case-insensitively only on Windows. APFS is case-insensitive by default but can be formatted case-sensitive, so macOS is treated as case-sensitive: comparing too strictly rejects a legitimate path, which is visible, while comparing too loosely admits an escape, which is silent.
- Rejects rooted input to `CombineWithin` outright, since `Path.Combine` discards its first argument when the second is absolute.
- Returns null when a path cannot be resolved, so the caller refuses it rather than comparing an unverified string.

Enumeration now sets `AttributesToSkip = FileAttributes.ReparsePoint` and re-confines each entry rather than trusting the one-time check at construction — the walk runs every sync cycle for the life of the process, so a directory swapped for a link in between would otherwise win (CWE-367).

## Not fixed, deliberately

Bind mounts and hard links are out of reach of any string check: the resolved path genuinely *is* beneath the root, so there is nothing left to inspect. Those need a low-privilege service account and a deployment that keeps the application's own config and key ring outside every configured root. Noted on #365 for the deployment docs.

The `allowedRoot` allowlist — constraining what the root itself may be, following Elasticsearch's static `path.repo` — is **not** in this PR. It belongs with the connection-configuration work in phase 3c of #351. The allowlist stops `allowedRoot: "/"`; it does not stop a symlink. Both controls are needed and they are independent.

## Testing

`dotnet test` — **1080 passed, 0 failed.** 13 new tests: 8 on the helper, 5 end-to-end on the connector, covering link escape, path-through-a-link, absolute-path bypass, `..` traversal, sibling-prefix, and that ordinary reads inside the root still work. Link-creation tests self-skip on platforms that will not create links unprivileged.

Background and prior art: `docs/research/programmatic-source-configuration-safety-2026-08-17.md`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened filesystem access controls to prevent directory traversal, absolute-path escapes, sibling-prefix bypasses, and symlink or junction-based escapes.
  * Filesystem listings now skip unsafe links and validate discovered paths before access.
  * Valid paths within the configured root continue to work as expected.

* **Documentation**
  * Added research and implementation plans covering source management, access restrictions, credential activation, MCP behavior, cloud storage boundaries, and infrastructure-as-code workflows.

* **Tests**
  * Added cross-platform coverage for path confinement, link handling, traversal rejection, and valid in-root access.

### DIFF
--- a/docs/research/programmatic-source-configuration-safety-2026-08-17.md
+++ b/docs/research/programmatic-source-configuration-safety-2026-08-17.md
@@ -1,0 +1,191 @@
+# How should Connapse safely allow external data-source configuration to be created programmatically?
+
+**Date:** 2026-08-17
+**Status:** Reviewed
+**Built on:** no prior corpus material (Connapse MCP was not available in this session; `docs/research/highest-value-connectors-2026-08-13.md` covers connector selection, not authorization)
+
+## Executive summary
+
+Connapse should expose source creation over REST behind an admin role, keep connection creation out of the API entirely, and put nothing that mutates configuration on MCP. Infrastructure-as-Code is preserved because the thing IaC needs to declare — sources — stays API-creatable; the credential boundary that IaC should *not* be able to invent stays file- and admin-only. This is the shape Grafana, Snowflake, Databricks, and Kubernetes all converged on independently: the object that names a credential is created through a lower-trust channel than the credential itself.
+
+Two server-side constraints must exist regardless of which surface is enabled. First, `allowedRoot` must be selected from a configured allowlist rather than supplied by a caller, matching Elasticsearch `path.repo`. Second — and this is a live defect in already-merged code, not a design question — Connapse's filesystem path confinement is **empirically bypassable by a directory junction**, verified during this research against the exact .NET APIs the product uses.
+
+The most significant correction to the team's prior thinking: AWS session policies, which looked like the elegant answer to bucket scoping, cannot be the primary control. They apply only when `roleArn` is set and have **no Azure equivalent**, so they cannot provide a uniform boundary across both clouds.
+
+## Research brief
+
+**Question:** How should a self-hosted RAG platform allow external data-source configuration to be created programmatically (REST/CLI, for IaC) without turning that capability into a security hole?
+
+**Sub-questions investigated:**
+1. What patterns do comparable self-hosted products use to reconcile IaC with programmatic data-source creation?
+2. Is "allowlist roots in config, select by reference via API" the established answer for filesystem-root authority scopes, and what are its failure modes?
+3. How do these systems separate "declare configuration" from "activate/grant authority"?
+4. What is best practice for exposing mutating tools on an MCP server given prompt-injection risk from ingested content?
+5. How do products constrain which buckets a scope may name, when the credential itself is broad?
+
+**Out of scope:** Connapse UI design; the Phase 4 Sources/Connections screens; regulatory compliance; multi-tenant SaaS tenancy models (Connapse is self-hosted, single-org).
+
+**Success criteria:** A per-surface recommendation (Admin UI / REST+CLI / MCP) for connections vs sources, the server-side constraints required regardless of surface, and an IaC story that does not weaken the boundary.
+
+## Findings by sub-question
+
+### 1. IaC patterns in comparable self-hosted products
+
+**Claim: the established pattern is channel-scoped authority — the file/config channel can create objects the HTTP API cannot.**
+
+Grafana is the clearest precedent and the closest analogue to Connapse. Datasources are provisioned from YAML in `provisioning/datasources/`, read at startup, with version control as the stated motivation. The security-relevant part is the asymmetry: a provisioned datasource carries `editable` (default false), surfacing as `readOnly`, and **`readOnly` cannot be set through the HTTP API at all**. Grafana PR #19006 ignores the field on create/update with the explicit rationale that "one should not be able to create a readonly data source via the API." The filesystem is treated as the high-trust channel because writing it already implies host access. [primary]
+
+That boundary has leaked in practice — Grafana issue #32556 reports a `PUT` on a provisioned datasource flipping `readOnly` back to false. Worth noting as evidence that this pattern needs test coverage, not merely intent. [primary, single issue]
+
+Grafana does expose `POST /api/datasources`, gated by a `datasources:create` permission. The OSS default role mapping for that permission was not confirmed. [primary for the permission name; unverified for the default role]
+
+Airbyte takes the opposite approach — full creation over its public API, with the guardrail in RBAC granularity rather than channel. It has a **Source editor** role distinct from Destination editor, explicitly so teams can start syncing "without being able to reconfigure the destinations they write to." Reading data and writing data are separate privileges. Its Terraform provider is Speakeasy-generated directly from the OpenAPI spec, adding no separate authority model. [primary]
+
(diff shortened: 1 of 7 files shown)


Destrayon:
Thanks — worked through all 9. Pushed in 2446e33. Six applied, two declined with reasons, one refuted.

## Applied

**The .NET version was wrong, and the catch is a good one.** I cited ".NET 7.6.4" — that was PowerShell's version read from `$PSVersionTable.PSVersion`, not the runtime's. The project targets `net10.0` and `dotnet --version` reports 10.0.303. Corrected in the research doc, this PR body, and #365. The reproduction itself is unaffected, but citing the wrong runtime in a security writeup is exactly the sort of thing that makes a reader distrust the rest of it.

**Case-sensitivity inference was wrong in the risky direction.** I had case-insensitive everywhere except Linux. APFS is case-insensitive *by default* but can be formatted case-sensitive, and on such a volume `/Data` and `/data` are genuinely different directories — matching them loosely admits one as inside the other. Now case-insensitive only on Windows, where NTFS is reliably case-insensitive and an ordinal compare would wrongly reject `C:\Data\x` as outside `C:\data`. The failure modes are not symmetric, which is what decides it: too strict rejects a legitimate path and is visible; too loose admits an escape and is silent.

**Dropped the sentinel.** `ResolveLinks` returned `Path.Combine(fullPath, "\0unresolvable")` and relied on that never matching a root. Correct in practice, fragile as a security check. It now returns null, `ResolveWithin` refuses on null, and it additionally catches the malformed input `Path.GetFullPath` rejects outright.

**Research doc accuracy** — the empirical table is now labelled as pre-fix and records what this PR changed, so it does not read as current behaviour. Added the missing primary citation for CVE-2021-43798 (NVD plus the Grafana advisory).

**Plan doc, phase 3c** — source mutations take `RequireAdmin` rather than `RequireEditor`; responses use DTOs that omit `ScopeJson`, with tests asserting its *absence* so a later refactor that serializes the record directly cannot pass silently; and mutating `/api/connections` routes are now recorded as decided out of scope rather than left open, with a note that reopening them reopens the security review.

## Declined

**Atomic no-follow open for the TOCTOU gap.** Correct in principle and I am not disputing the analysis — validate-then-open is a genuine race. But .NET has no cross-platform anchored-open primitive; doing this properly means `openat2` with `RESOLVE_BENEATH` on Linux and per-platform P/Invoke elsewhere. That is a substantial piece of work with its own risk, and it does not belong bolted onto a fix that is otherwise self-contained. It is already documented as a residual in the research doc and on #365, alongside bind mounts and hard links, which no amount of path checking can address. Exploiting it needs local write access inside an already-configured root plus a win on the race.

**Async `TryCreateDirectoryLink`.** It is a test helper spawning `mklink`, which emits a single line. The deadlock this pattern can cause needs a filled output buffer, which will not happen here, and `WaitForExit(10_000)` already bounds it. Converting three synchronous tests to async to remove a blocking call in test setup is churn without a behavioural change.

## Refuted

**Integration test for deleting a referenced connection** — this already exists: `Connections_DeleteWithReferencingSource_IsRestricted` in `tests/Connapse.Integration.Tests/OwnerBridgeSchemaTests.cs:157`. It asserts a `ForeignKeyViolation` on deleting a connection that a source still references.

`dotnet test` — **1080 passed, 0 failed.**

PR #367, opened 2026-08-19
feat(storage): bound what a source may be pointed at

Destrayon:
Part of #351.

Bounds what a source may be pointed at, on both sides: a configured allowlist of filesystem roots, and a per-connection allowlist of cloud storage locations. Together these are the precondition for exposing source creation over REST at all — without them, "create a source" is "read anything the deployment's credentials can reach, and make it searchable."

Background and prior art: `docs/research/programmatic-source-configuration-safety-2026-08-17.md`.

## `Sources:Security:AllowedFilesystemRoots` — deployment configuration

A connection's `allowedRoot` must equal a configured entry or sit beneath it. Configuration only, never the settings table: the authority a filesystem root confers is the same class of thing as a cloud credential, and this project already refuses to accept those over an API. Elasticsearch takes the same line with `path.repo`, a static node setting its API cannot change.

This is what stops `allowedRoot: "/"`. It is independent of the link resolution in #366 and neither substitutes for the other — an allowlist does not stop a symlink, and resolving links does not stop a root pointing at the whole disk. Root matching resolves links on both sides, so `/data/link-to-etc` cannot pass by appearing to sit under a permitted `/data`.

## `allowedLocations` — per-connection, for S3 and Azure Blob

A source's bucket or blob container plus prefix must fall inside one of the entries its connection declares.

**IAM cannot make this distinction.** One connection role is shared by every source using that connection, so to AWS or Azure they are the same principal. Snowflake reaches the same conclusion with `STORAGE_ALLOWED_LOCATIONS` on a storage integration — shipped alongside narrow IAM, not instead of it. Databricks Unity Catalog splits the same way, separating a storage credential from external locations.

Worth recording why this is not done with AWS session policies, which looked like the more elegant answer: they apply only when `roleArn` is set, cap at 2 KB, and **have no Azure equivalent** — managed-identity tokens cannot be narrowed at request time. They cannot provide a uniform boundary across both clouds. Still worth adding opportunistically on the assume-role path later, where they would convert an application bug into an AWS-side denial.

Matching rules: an entry naming only a bucket permits any prefix within it; one naming a bucket and prefix permits only that subtree. Comparison is case-sensitive, because S3 bucket names and Azure container names both are, and a separator is appended before prefix-matching so `bucket/docs` cannot admit `bucket/docs-internal` — the off-by-slash trap that breaks prefix comparisons.

## Both enforce at the factory

`ConnectorFactory.Create(source, connection)` is the single point every source connector is built through, so a future endpoint that forgets to validate cannot bypass either check.

## Upgrade behaviour — deliberately permissive for one release

An empty allowlist permits and logs a warning naming the root or container in use. #350 backfilled existing Filesystem, S3, and Azure containers into connections, and none of them declare either allowlist, so enforcing immediately would break every upgrade until an operator edited configuration. The warning tells them what to add. Both become deny-by-default in a later release.

This is a deliberate softening of the research's "absent means deny" recommendation. It is defensible because connections remain Admin-UI-only, so an unconstrained root or location is not remotely reachable today — but it must be tightened before, or together with, any programmatic connection creation.

## Testing

`dotnet test` — **1099 passed, 0 failed.** 16 new tests: 12 on the location policy (bucket-only entries, prefixed subtrees, sibling-name rejection on both bucket and prefix, case sensitivity, slash normalization, and that a list of blank entries does not read as "unconfigured"), and 4 at the factory covering both clouds plus the grace path.

## Summary by CodeRabbit

* **New Features**
  * Added configurable security restrictions for filesystem roots used by sources.
  * Added storage-location allowlists for cloud containers, buckets, and prefixes.
  * Added clear allowed, denied, and unrestricted validation outcomes.
  * Unrestricted configurations remain supported with warning notifications, issued only once per connection and location.

* **Configuration**
  * Added the `Sources.Security` section for managing permitted filesystem roots.

* **Tests**
  * Added coverage for location matching, path boundaries, normalization, case sensitivity, warning behavior, and security enforcement.

### DIFF
--- a/src/Connapse.Core/Models/SourceSecuritySettings.cs
+++ b/src/Connapse.Core/Models/SourceSecuritySettings.cs
@@ -12,7 +12,7 @@ namespace Connapse.Core;
 /// <c>path.repo</c>, which is a static node setting requiring a restart to change.
 /// </para>
 /// </summary>
-public class SourceSecuritySettings
+public record SourceSecuritySettings
 {
     public const string SectionName = "Sources:Security";
 
--- a/src/Connapse.Core/Models/SourceSecuritySettings.cs
+++ b/src/Connapse.Core/Models/SourceSecuritySettings.cs
@@ -1,0 +1,66 @@
+using Connapse.Core.Utilities;
+
+namespace Connapse.Core;
+
+/// <summary>
+/// Bounds what external data a source may be pointed at, independent of who points it.
+/// <para>
+/// Deliberately configuration-only — appsettings, environment, or the deployment's own
+/// config file — and never writable through the API or the settings table. The authority a
+/// filesystem root confers is the same class of thing as a cloud credential, and this project
+/// already refuses to accept those over an API. Elasticsearch takes the same line with
+/// <c>path.repo</c>, which is a static node setting requiring a restart to change.
+/// </para>
+/// </summary>
+public class SourceSecuritySettings
+{
+    public const string SectionName = "Sources:Security";
+
+    /// <summary>
+    /// Directories a filesystem connection's <c>allowedRoot</c> may name. A root must equal
+    /// one of these or sit beneath it.
+    /// <para>
+    /// Empty means unrestricted, which is the pre-existing behaviour and is retained for one
+    /// release so upgrades do not break every filesystem source that #350 backfilled. The
+    /// unrestricted case logs a warning naming each root in use, so an operator can see what
+    /// to configure before it becomes deny-by-default.
+    /// </para>
+    /// </summary>
+    public IReadOnlyList<string> AllowedFilesystemRoots { get; set; } = [];
+
+    /// <summary>
+    /// Decides whether a filesystem root is permitted, resolving links on both sides so a
+    /// symlinked root cannot masquerade as an allowed one.
+    /// </summary>
+    public FilesystemRootDecision EvaluateRoot(string allowedRoot)
+    {
+        ArgumentException.ThrowIfNullOrWhiteSpace(allowedRoot);
+
+        if (AllowedFilesystemRoots.Count == 0)
+            return FilesystemRootDecision.UnrestrictedByConfiguration;
+
+        foreach (string permitted in AllowedFilesystemRoots)
+        {
+            if (string.IsNullOrWhiteSpace(permitted)) continue;
+
+            // ResolveWithin, not a string compare: a root of "/data/link-to-etc" would
+            // otherwise pass by looking like it sits under an allowed "/data".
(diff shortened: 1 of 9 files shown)


Destrayon:
All seven applied, none declined. Pushed in b386451. Two were more than nitpicks.

## The actionable one was a real hole, and my own test hid it

A list of blank `allowedLocations` entries fell through to `UnrestrictedByConfiguration`, so a typo — `allowedLocations: ["", ""]` — silently permitted every bucket the credential could reach. The distinction now enforced is between an **absent** control and a **broken** one: only a genuinely empty list takes the grace path; a declared-but-blank list denies.

Worth being explicit about how this got in, because it is the more useful part of the finding. I named the test `Evaluate_BlankEntriesAreIgnoredButDoNotMakeItUnrestricted` and wrote a comment saying "a list of blanks must not read as nothing configured and silently permit everything — that would turn a typo into an open door" — then asserted `UnrestrictedByConfiguration` directly underneath. The intent was right in prose and inverted in code, and the test passing made it look verified. Now split into two tests that each assert one thing.

## The warning flood was a design flaw, not a nitpick

Correct that `Create(Source, Connection)` runs per sync cycle. An unconfigured deployment would have emitted one warning per source every five minutes, which defeats the purpose — the grace period only works if the warning is *noticed*, and a flood guarantees it is not.

Now warned once per connection-and-scope, keyed on both so that changing a root or bucket warns again rather than staying silent behind an earlier entry. Two tests cover it: repeated cycles warn once, and two scopes on one connection each warn.

## The rest

- **Grace warnings are now asserted.** You were right that `NullLogger` meant nothing verified the one signal an operator gets. Used a small recording logger rather than a substitute — `ILogger.Log` being generic makes the mock assertion harder to read than the thing it checks. The assertions check content, not just level: the cloud warning must name the container, the filesystem one must name `AllowedFilesystemRoots`.
- **`SourceSecuritySettings` is a record**, per the settings-model convention. I added a configuration-binding test alongside it, which was not requested but seemed worth having: this allowlist is *only* reachable through binding, and a silent binding failure would be indistinguishable from a legitimately unconfigured deployment — the security control would simply be off with no symptom.
- **Primary constructor** on `ConnectorFactory`, per the DI convention.
- **Dead `EndsWith` branch removed** — correct, `Normalize` trims trailing slashes so it was unreachable. The boundary behaviour it guarded is unchanged and still covered.
- **Test renamed** to `Evaluate_LocationDifferingOnlyByCase_IsDenied`.

`dotnet test` — **1108 passed, 0 failed** (up from 1099; 9 new tests).

PR #368, opened 2026-08-20
feat(web): add /api/sources with admin-scoped mutations

Destrayon:
Part of #351. Task 1 of phase 3c.

Sources have been syncable and searchable since #364 but reachable only through the database. This adds `GET /api/sources`, `GET /api/sources/{id}`, `POST`, `PATCH`, `DELETE`, and `POST /api/sources/{id}/sync`.

## What is deliberately absent

**Nothing returns documents or paths.** No browse route, no document listing. A source is a searchable *scope*; enumerating one exposes every synced object to any Connapse reader regardless of what they could see in the source system — the leak epic #348 exists to close. A test pins that `/documents`, `/browse`, and `/files` stay 404, and another that a source id on the container document route is refused rather than quietly enumerating through the other surface.

**Connections have no REST surface.** They are the credential and filesystem-root boundary, configured out of band and managed in Admin Settings. Rationale and prior art: `docs/research/programmatic-source-configuration-safety-2026-08-17.md`.

## Two decisions that differ from the container endpoints

**Mutations require `RequireAdmin`, not `RequireEditor`.** Creating a source chooses what external data gets indexed and made searchable, bounded only by whatever the connection's credential can reach. That is an administrative act, not a content one. Airbyte separates its source-editor role from destination-editor for the same reason. Copying the container endpoints' authorization here would have been the easy mistake, so the tests assert an Editor is *forbidden* rather than only that an Admin succeeds.

**Responses use a `SourceResponse` DTO, never the `Source` record.** The record carries `ScopeJson` — naming buckets, prefixes, and filesystem subpaths — and `SyncCursor`, an opaque provider continuation token. Serializing it directly hands both to any reader. The tests assert their absence against the **raw response body** rather than a deserialized DTO, because deserializing into `SourceResponse` would silently drop an unexpected field and pass. Positive assertions run first so an empty or error body cannot satisfy the negatives.

## `LastSyncError` is administrator-only

Worth calling out because it is not obvious. Reads are open to viewers, but a provider's failure text routinely quotes the resource that failed — `Access Denied for bucket payroll-secrets` — so returning it to every reader would give back exactly the infrastructure detail `ScopeJson` is withheld to protect. The test asserts a viewer does not see the bucket name and an administrator does.

## One reversal

`SourceSyncService` returns to singleton registration so the sync-now route can drive the same instance. #364 registered it as a plain hosted service on the explicit grounds that nothing called into it at runtime; this route is that caller, so the reasoning no longer holds.

## Testing

`dotnet test` — **1110 passed, 0 failed.** 15 new integration tests covering scope omission on both list and detail, the viewer/admin split on sync errors, Editor being forbidden from create and delete, Viewer being allowed to list, unknown-connection and malformed-scope rejection, duplicate-name conflict, update, delete, sync-when-disabled, and the absent enumeration routes.

Still to come in phase 3c: the MCP `kind` discriminator, which needs a second id resolver so `list_files` stays container-only.

## Summary by CodeRabbit

* **New Features**
  * Added API endpoints to list, view, create, update, delete, and manually sync sources.
  * Added source configuration options for scope, description, and sync intervals.
  * Added role-based access controls, with diagnostic details restricted to administrators.
  * Added validation for connections, duplicate names, malformed settings, and disabled-source syncing.

* **Bug Fixes**
  * Prevented source scope, cursors, and documents from being exposed through read responses.
  * Prevented overlapping syncs for the same source and now return a conflict when one is already running.

### DIFF
--- a/src/Connapse.Core/Models/SyncModels.cs
+++ b/src/Connapse.Core/Models/SyncModels.cs
@@ -21,12 +21,19 @@ public record OwnerRef(Guid Id, bool IsSource)
 /// <summary>
 /// Outcome of one sync cycle for one source.
 /// </summary>
+/// <param name="AlreadyRunning">
+/// True when the cycle did nothing because another sync of the same source held the gate.
+/// Distinct from <paramref name="Error"/>: nothing went wrong, the work simply belongs to
+/// the cycle already in flight. A caller that surfaces this to a user should say "in
+/// progress", not "failed".
+/// </param>
 public record SourceSyncResult(
     int Upserted,
     int Deleted,
     bool UsedDeltaPath,
     bool RequiredResync,
-    string? Error);
+    string? Error,
+    bool AlreadyRunning = false);
 
 /// <summary>
 /// Thrown when a reindex would move a document between ownership domains — source to
--- a/src/Connapse.Web/Endpoints/SourceResponse.cs
+++ b/src/Connapse.Web/Endpoints/SourceResponse.cs
@@ -1,0 +1,55 @@
+using Connapse.Core;
+
+namespace Connapse.Web.Endpoints;
+
+/// <summary>
+/// What the API returns for a source.
+/// <para>
+/// Deliberately not the <see cref="Source"/> record. That carries <c>ScopeJson</c>, which
+/// names buckets, prefixes and filesystem subpaths, and <c>SyncCursor</c>, an opaque
+/// provider continuation token — infrastructure detail a reader has no reason to receive,
+/// and the kind of thing that turns a read route into reconnaissance. Serializing the record
+/// directly would hand both out.
+/// </para>
+/// </summary>
+public record SourceResponse(
+    Guid Id,
+    string Name,
+    string? Description,
+    Guid ConnectionId,
+    bool Enabled,
+    SyncStatus LastSyncStatus,
+    DateTime? LastSyncedAt,
+    int? SyncIntervalSeconds,
+    int DocumentCount,
+    string? Summary,
+    DateTime CreatedAt,
+    DateTime UpdatedAt,
+    /// <summary>
+    /// Populated for administrators only. A provider's failure text routinely echoes the
+    /// thing that failed — "Access Denied for bucket payroll-data" — so returning it to
+    /// every reader would give back exactly the infrastructure detail ScopeJson is withheld
+    /// to protect.
+    /// </summary>
+    string? LastSyncError,
+    /// <summary>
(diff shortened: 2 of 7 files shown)


Destrayon:
Three applied, one refuted. Pushed in 9bf7e13.

## Applied — the concurrency finding was the real one

Correct, and it is a hazard **this PR introduced**. The timer was previously the only caller of `SyncSourceAsync` and iterates its sources sequentially, so two cycles for one source could not overlap. The sync-now route makes that possible.

The compare-and-swap on the cursor keeps progress safe either way — that was built in #364 — so this is not a correctness bug. But two cycles list the same remote twice and enqueue the same files twice, which is duplicated embedding work and costs real money.

`SyncSourceAsync` now takes a per-source gate, acquired **without waiting** so a caller arriving mid-cycle hears "already running" instead of blocking behind a five-minute sync. Different sources still run concurrently. `SourceSyncResult` gains `AlreadyRunning`, kept distinct from `Error` because nothing went wrong, and the endpoint returns **409** rather than 200 — a script driving this would otherwise read "0 upserted" as "nothing to do".

Worth flagging how I tested it. My first version raced two HTTP requests and asserted one got a 409. It passed, but only because the timing happened to work — that test would pass or fail on load. Replaced with a connector that blocks inside its remote call, so the interleaving is deterministic, plus two tests that the gate is **released** after a successful and a failed cycle. A missed release would wedge a source permanently, and the failure path is where it would hurt most, since an unreachable remote is exactly when cycles repeat.

**`JsonDocument` disposal** — right, it rents pooled buffers and this runs per request. Now `using` inside both try blocks.

**Test rename** — done, `GetSource_EnumerationRoutes_AreNotFound`.

## Refuted — `LogSanitizer` on the audit payloads

`IAuditLogger.LogAsync` serializes `details` with `JsonSerializer` and stores the result in the `Details` **jsonb column** of `audit_logs`. It never passes that object to `ILogger`; the only logger call in `AuditLogger` is on failure and logs the hardcoded `action` string, never the payload.

Log forging is an attack on log *text* — injecting newlines to forge entries. A value bound as a JSON property in a database column is not concatenated into a log line, and the serializer escapes control characters regardless. The repository guideline covers logging user-controlled values, which this is not.

The existing `ContainersEndpoints` does the same thing — `new { container.Name, container.ConnectorType }` with no sanitizer — so applying it here would have made these four call sites inconsistent with every other audit call in the codebase without closing anything.

Happy to reconsider if CodeQL actually flags it, since that would be evidence I have the data flow wrong.

`dotnet test` — **1098 passed, 0 failed.**

PR #371, opened 2026-08-21
feat(web): add the Connections settings tab

Destrayon:
Part of #352. Task 4a of phase 4. Plan: `docs/superpowers/plans/2026-08-18-connector-source-split-phase-4.md`.

Connections could previously only be created by writing to the database. This adds the tab where they are managed — and deliberately the only place: there is no REST, CLI, or MCP route that creates or edits one, because the authority a connection confers is the same class of thing as a cloud credential, which this project does not accept over an API.

## It looks like its siblings but isn't built like them

Every other settings tab edits one strongly-typed record through `IOptionsMonitor` and hands it back via an `OnSave` callback. This one is entity CRUD over `IConnectionStore`, so it loads and saves itself and takes no parameters. Matching the visual pattern without the data pattern is deliberate, but it will look inconsistent next to the others on first read.

## Three constraints, each carried from the design research

**The filesystem root is selected, not typed.** When `Sources:Security:AllowedFilesystemRoots` is configured, it renders as a dropdown of those roots. The whole point of that allowlist is that the interface *chooses among* configured roots rather than inventing one — a free-text box here would reopen exactly the hole it closes. With no allowlist configured it falls back to a text field with a visible warning, matching the one-release grace window the factory already implements.

**The secret is never rendered.** `Connection` carries `HasSecret`, not the secret, and the editor never populates that field from storage — it only ever holds a replacement the operator just typed. Blank means keep. There is no read path for a stored secret outside the sync engine and the UI must not become one.

**Deleting a referenced connection explains itself.** `SourceCount` is already on the model, so it says "3 sources use this connection" rather than surfacing the store's exception.

## One deviation from the issue

The issue says to reuse `S3ConnectionTester` and `AzureBlobConnectionTester` unchanged. They can't test a connection as written: both require a bucket or blob container (`Missing BucketName`, `Missing ContainerName`), and that name belongs to a **source**, not a connection.

Resolved without touching either tester — the Test button takes a probe target, defaulting to the first declared allowed location, since that is by definition somewhere this connection is meant to be able to reach. When none is declared, a small field asks which bucket to probe. The probe target is never persisted; it exists only to give the tester the argument it needs.

## Why the mapping is a separate class

`ConnectionForm` holds the form↔JSON mapping as a plain class rather than logic inside the Razor component. This repository has no component test harness — no bUnit, no existing component tests — and adding one would be a larger change than the feature. The mapping is also the part that carries real risk: a field dropped in the round trip produces a connection that resolves somewhere other than the operator intended, and nothing downstream can tell the difference.

20 tests cover it: the round trip per provider, that switching provider cannot leave a stale key the connector factory would honour, that an absent `allowedLocations` stays absent rather than becoming a deny-everything empty array, that the managed-identity client id survives (dropping it silently falls back to the default identity, which may have wider access), and that a secret never surfaces even when present in stored JSON.

## Testing

`dotnet test` — **1155 passed, 0 failed.**

The tab itself is not covered by automated tests, for the reason above. Worth a manual pass before merge: create one connection per provider, confirm the root is a dropdown when an allowlist is configured, confirm the secret field stays blank when editing a connection that has one, and confirm delete is refused while a source references it.

## Next

4b adds the Sources tab, 4c simplifies Home and removes the Filesystem write flags. #352 closes with 4c.

## Summary by CodeRabbit

- **New Features**
  - Added a Connections tab for managing filesystem, S3, and Azure Blob connections.
  - Added connection creation, editing, deletion, validation, and connectivity testing.
  - Added secure credential status indicators and replacement-only secret handling.
  - Added provider-specific location restrictions and filesystem root selection.
  - Added connection forms that handle invalid or incomplete configuration safely.

- **Tests**
  - Added coverage for connection form validation, serialization, location parsing, and secret handling.

- **Documentation**
  - Added planning documentation for the next connector and source management phase.

### DIFF
--- a/docs/superpowers/plans/2026-08-18-connector-source-split-phase-4.md
+++ b/docs/superpowers/plans/2026-08-18-connector-source-split-phase-4.md
@@ -1,0 +1,104 @@
+# Phase 4 — Sources tab, Connections settings, Home simplification
+
+Closes #352. Part of epic #348. Follows Phase 3 (#351, merged).
+
+## Why this phase is the one that matters
+
+Phases 1–3 built the machinery: schema, ownership, sync engine, REST surface, MCP contract,
+and the security work around all of it. None of it changed what a user sees. All three
+problems the epic exists to fix are **presentation** problems:
+
+1. Rendering an S3 bucket as a browsable folder tree implies ownership Connapse does not have.
+2. The file tree exposes every synced object to any Connapse user, regardless of what they
+   could see in the source system. Browsing is a far wider surface than search results, which
+   already pass through `CloudScopeService`.
+3. Synced buckets compete with real managed containers in one list.
+
+Until the file browser stops serving external connectors, all three persist exactly as they
+did before the epic started.
+
+**Done when:** no external connector is reachable through the file browser.
+
+## Task 4a — Connections settings tab
+
+**Goal:** an administrator can register and test a connection without touching the database.
+
+New `ConnectionsSettingsTab.razor`, added to the `Settings.razor` tab host. It differs from
+every existing tab in kind: the others edit one settings record through `IOptionsMonitor` and
+`OnSave`, while this one is entity CRUD over `IConnectionStore`. Follow the visual pattern,
+not the data pattern.
+
+Per provider: Filesystem takes a root; S3 takes region and an optional role ARN; Azure Blob
+takes a storage account and an optional managed-identity client id. All three take an optional
+`allowedLocations` list.
+
+Three constraints that are not negotiable, each with a reason recorded elsewhere:
+
+- **`allowedRoot` is selected, not typed.** When `Sources:Security:AllowedFilesystemRoots` is
+  configured, render a dropdown of those roots — the API's job is to *select among* configured
+  roots, never to invent one. When the allowlist is empty (the one-release grace window),
+  fall back to a text field with a visible warning that it will become deny-by-default.
+- **The secret is never displayed.** `Connection` carries `HasSecret`, not the secret. Show
+  "configured" or "not configured" and offer replacement only. There is no read path for a
+  stored secret outside the sync engine, and the UI must not become one.
+- **Deletion explains itself.** `IConnectionStore.DeleteAsync` refuses while sources still
+  reference the connection. `Connection.SourceCount` is already on the model — say "3 sources
+  use this connection" rather than surfacing a raw exception.
+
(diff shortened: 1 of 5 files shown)


Destrayon:
Nine applied, one refuted. Pushed in cbfe9a8. One of them turned into a removal rather than a fix.

## The copy finding was right, and the fix was to delete the feature

You flagged "Connapse never stores cloud keys" as contradicted by the credential field underneath it. Correct — but the interesting part was checking *who consumes that value*.

`GetSecretAsync` has **no caller anywhere** outside its own store: not `ConnectorFactory`, not `SourceSyncService`, nothing. No connector reads a connection secret. The field was dead weight, and worse, it invited an operator to paste an AWS access key — the exact thing this project refuses to hold, with the environment-based credential model existing specifically to avoid it.

So the field is gone rather than reworded, and the copy is now true instead of aspirational. The schema keeps its secret columns (Phase 1 built them, and a future provider may need one); the UI simply never writes one. A row that already carries a secret shows as a **legacy secret** in the list rather than being hidden, so an operator can see it is still there.

## The parsing bug was real and would have taken the tab down

`GetValue<string>()` on `allowedLocations` entries throws on a non-string, and it sits *outside* the parse guard — so one numeric element in a stored config would have killed the whole tab, not just that field. Now skips non-strings, with a test covering `["good", 42, null, {"x":1}, "also-good"]`.

## The concurrency findings

Both correct. Save reported success when `UpdateAsync` returned null, and delete when `DeleteAsync` returned false — each meaning the row was removed from another session. Both now say so and reload rather than lying.

Delete also caught only `InvalidOperationException`, so a database error would escape the Blazor event handler and tear down the circuit — the operator would watch the page die instead of reading a message. The catch is deliberately broad now.

`ConnectionForm` is a record per the DTO convention.

## Plan updates

All four applied. The most substantive: 4c now specifies the test that makes "no external connector reachable through the file browser" an enforced property rather than a stated goal — seed a source and a container, drive the route for each, and require the source to be **refused** rather than return an empty listing. An empty result passes today and silently becomes a real listing the first time someone widens the lookup, which is precisely the regression the phase exists to prevent.

I also corrected the plan's own secret bullet, which still described the write-only field that no longer exists.

## Refuted

**File-scoped namespaces** — both files already use them. `ConnectionForm.cs:4` is `namespace Connapse.Web.Components.Settings;` and `ConnectionFormTests.cs:7` is `namespace Connapse.Core.Tests.Settings;`, both with the semicolon form and no wrapping braces.

`dotnet test` — **1156 passed, 0 failed.**

PR #372, opened 2026-08-21
feat(web): add the Sources tab

### LINKED ISSUE #352: Phase 4 — Sources tab, Connections settings, and Home simplification

Destrayon:
Part of #352. Task 4b of phase 4. Independent of #371 — branched off `main`, touches no file that PR touches.

Sources have been syncable since #364 and manageable over REST since #368. This is the first place an operator can see them in the product.

## What the page deliberately does not have

No file list, no path, no document name. That absence *is* the feature: the file tree exposed every synced object to any Connapse user regardless of what they could see in the source system, which is a far wider surface than search results — and search already passes through `CloudScopeService`.

The aggregate document count is the one exception, accepted during design as a weak signal in exchange for the page being useful at all.

## Authorization

Reads are open to any authenticated user; **sync, enable/disable, and delete are admin-only**, matching `/api/sources`.

The last sync error is admin-only too. A provider's failure text routinely quotes the resource that failed — `Access Denied for bucket payroll-secrets` — which is the same infrastructure detail the API withholds from non-admin readers. Showing it to everyone here would undo that.

## Scope rendering is a separate class

`SourceScopeSummary` renders the one-line "where does this point" cell. It is a separate class rather than logic inside the component so it can be tested — this repository has no component test harness, and adding one would be a larger change than the feature.

It reads **only** the keys that name the scope (`bucketName`, `containerName`, `prefix`, `subPath`, and the patterns). A test pins that a stored config carrying stray keys like `documents` or `lastFile` cannot put document detail on the page — the Sources tab showing file names is precisely the leak this epic exists to close, and it would be an easy one to introduce later by rendering the scope blob more generously.

It also skips non-string pattern entries rather than throwing: a display helper that throws takes the whole page down over cosmetic detail.

## Two smaller decisions

**Sources sits beside Files in the nav**, not under Settings. Managed storage and mirrored external systems are peers, and that split is exactly what the navigation should make legible. Connections stay in Settings, since those are configuration rather than content.

**Sync-now duplicates the endpoint's sequence** rather than calling the REST API from the server. Blazor Server and REST are two genuine front doors, and having the page authenticate against its own API would be worse. It shares the per-source gate added in #368, so a manual sync during a scheduled cycle reports "already in progress" rather than starting a second one.

## Testing

`dotnet test` — **1149 passed, 0 failed.** 14 new tests on the scope summarizer.

The page itself has no automated coverage, for the harness reason above. Worth a manual pass: confirm a viewer sees no action buttons and no error text, confirm sync-now reports rather than throws when the remote is unreachable, and confirm the delete dialog states the external data is untouched.

## Next

4c is the one that closes #352: Home simplification, removing the Filesystem write flags, the search facet — and the test that makes "no external connector reachable through the file browser" an enforced property rather than a stated goal.

## Summary by CodeRabbit

* **New Features**
  * Added a Sources page for viewing external source details, scope summaries, document counts, and synchronization status.
  * Administrators can sync, enable or disable, and delete sources with confirmation and status feedback.
  * Added a Sources link to the navigation menu.
  * Non-administrators can view source information in read-only mode.
  * Added clear summaries for storage locations, paths, inclusion patterns, and exclusions.

* **Bug Fixes**
  * Improved handling of invalid or incomplete source scope data without displaying errors or document details.

### DIFF
--- a/src/Connapse.Web/Components/Layout/NavMenu.razor
+++ b/src/Connapse.Web/Components/Layout/NavMenu.razor
@@ -18,6 +18,15 @@
                     </NavLink>
                 </div>
 
+                @* Sources sits beside Files rather than under Settings: managed storage and
+                   mirrored external systems are peers, and the split between them is the
+                   distinction this nav is meant to make legible. *@
+                <div class="nav-item px-3">
+                    <NavLink class="nav-link" href="sources">
+                        <span class="nav-icon bi-cloud-arrow-down" aria-hidden="true"></span> Sources
+                    </NavLink>
+                </div>
+
                 <div class="nav-item px-3">
                     <NavLink class="nav-link" href="search">
                         <span class="nav-icon bi-search" aria-hidden="true"></span> Search
--- a/src/Connapse.Web/Components/Pages/Sources.razor
+++ b/src/Connapse.Web/Components/Pages/Sources.razor
@@ -1,0 +1,370 @@
+@page "/sources"
+@rendermode InteractiveServer
+@attribute [Authorize]
+@using Connapse.Core
+@using Connapse.Core.Interfaces
+
+@using Connapse.Web.Services
+@inject ISourceStore SourceStore
+@inject IConnectionStore ConnectionStore
+@inject SourceSyncService SyncService
+@inject IAuditLogger AuditLogger
+@inject AuthenticationStateProvider AuthStateProvider
+
+@*
+    Sources — external scopes Connapse mirrors but does not own.
+
+    This page shows what each source is *doing*, never what is *inside* it. There is no file
+    list, no path, and no document name anywhere on it, and that is the point of the whole
+    epic rather than an omission: the file tree exposed every synced object to any Connapse
+    user regardless of what they could see in the source system, which is a far wider surface
+    than search results, and search already passes through CloudScopeService.
+
+    The aggregate document count is the one exception, accepted during design as a weak signal
+    in exchange for the page being useful at all.
+*@
+
+<PageTitle>Sources - Connapse</PageTitle>
+
+<div class="container-fluid py-4">
+    <div class="d-flex justify-content-between align-items-start mb-4">
+        <div>
+            <h1 class="mb-1"><span class="bi-cloud-arrow-down me-2"></span>Sources</h1>
+            <p class="text-muted mb-0">
+                External systems Connapse keeps in step with. Their contents are searchable but not
+                browsable — Connapse mirrors this data, it does not own it.
+            </p>
+        </div>
+        @if (isAdmin)
+        {
+            <a class="btn btn-outline-secondary" href="/settings">
+                <span class="bi-gear me-1"></span> Connections
+            </a>
+        }
+    </div>
+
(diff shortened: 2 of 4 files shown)


Destrayon:
One applied, one refuted. Pushed in 2084842.

## Applied — and it was worse than one bug

`LoadAsync` let a store failure escape, with two distinct consequences:

- **Uncaught from `OnInitializedAsync`**, it tears down the Blazor circuit. The operator watches the page die rather than reading an error.
- **With no `finally`**, a throw left `loading` stuck at `true` permanently — so even after a later action succeeded, the page kept showing its spinner. That one survives the error rather than accompanying it, which makes it the nastier half.

Now reports the failure, empties the lists so nothing stale renders, and always clears the flag.

## Refuted — `LogSanitizer` on the audit payloads

Same finding as #368, and the same evidence still holds.

`IAuditLogger.LogAsync` serializes `details` with `JsonSerializer` and stores the result in the **jsonb `Details` column** of `audit_logs`. It never passes that object to `ILogger`; the only logger call in `AuditLogger` is on write failure and records the hardcoded `action` string, never the payload.

Log forging is an attack on log *text* — injecting newlines to fabricate entries. A value bound as a JSON property in a database column is not concatenated into a log line, and the serializer escapes control characters regardless. The repository guideline covers logging user-controlled values, which this is not.

Every existing audit call in the codebase omits the sanitizer for the same reason — `ContainersEndpoints` passes `new { container.Name, container.ConnectorType }` unsanitized. Applying it to these three call sites would make them inconsistent with every other audit call while closing nothing.

Still happy to reconsider if CodeQL flags it in practice, since that would mean I have the data flow wrong.

`dotnet test` — **1149 passed, 0 failed.**

Issue #353, opened 2026-08-14
Phase 5 — Remove compatibility shims and drop connector columns


Destrayon:
Part of #348. Blocked by #352. Closes the epic.

## What

Remove the Phase 2 compatibility read path and drop `containers.connector_type` and `containers.connector_config`.

## Why

The compatibility window — where `containers` and `sources` both hold external-connector rows and the compat read must pick correctly — is the one genuinely awkward state in this design. It is scheduled into **the same milestone as the rest of #348** so it spans one release, not several. Leaving it open-ended is how it becomes permanent.

## How

Single PR, ~100 lines.

- Delete the compat read from the container endpoints.
- Migration dropping the two columns. Every remaining `containers` row is managed storage, so the columns no longer carry information.
- The `ConnectorType` enum itself **survives** — connections still need it as their `Provider`, minus `ManagedStorage`. Only the columns go.

## Done when

The compat read path no longer exists and no code references `containers.connector_type`.

PR #373, opened 2026-08-21
feat(web): stop serving external connectors from the file browser

### LINKED ISSUE #352: Phase 4 — Sources tab, Connections settings, and Home simplification

Destrayon:
Closes #352. Task 4c — and the phase's done-condition: **no external connector is reachable through the file browser.**

Phases 1–3 built the machinery without changing anything a user sees. All three problems this epic exists to fix are presentation problems, and this is where they stop:

1. Rendering an S3 bucket as a browsable folder tree implied ownership Connapse does not have.
2. The file tree exposed every synced object to any Connapse user regardless of what they could see in the source system — a far wider surface than search results, which already pass through `CloudScopeService`.
3. Synced buckets competed with real managed containers in one list.

**188 insertions, 584 deletions.** Most of this PR is removal.

## Home

The create-container modal drops the connector picker and every provider config block. `ContainersEndpoints` has rejected non-managed `ConnectorType` since 3a, so the form was offering choices the server refuses.

The folder browser and the Test Connection handler went with them — both existed only to fill fields that no longer exist, and connection testing now lives in the Connections tab from #371.

## FileBrowser

`IsReadOnly` and the three Filesystem permission flags are gone. Worth being precise about why that was safe: once containers became managed-only, **every one of them evaluated to a constant**, so the guards collapsed to the role check already sitting beside them. Nothing changed behaviour; the conditions were already dead.

The flags are removed from `FilesystemConnectorConfig` too. They were UI hints rather than enforcement in any case — write capability has been a type guarantee since 3a, where only `MinioConnector` implements `IWritableConnector`.

## `FileBrowserChangeNotifier` deleted

It has had no publisher since #364 removed the watcher, and the sources UI does not want one: a source has no file list to keep live. Leaving a dead event bus in place would only invite someone to wire it back up to the thing this epic just disconnected.

## Search

Gains a source facet, **grouped rather than merged** into one flat list — which kind a result came from changes how far it can be trusted to be current, so collapsing the distinction in the scope picker would undo it.

Auto-select now fires only when exactly one scope exists in total. Previously it picked the first container; with sources present that would have silently excluded everything Connapse mirrors, which is a bad default to inherit.

## The boundary is asserted, not stated

`FileBrowserBoundaryTests` pins the property this whole epic is about. Two things about how it is written:

**Every assertion is paired against the same route** — a container browses, a source is refused. Checking only the refusal would pass on a server that had broken browsing entirely; checking only the container would pass on one that browsed everything. The property is the *difference* between them, so a single test asserts both halves.

**Refusal, not emptiness.** An empty listing satisfies "no files were shown" today and silently becomes a real listing the first time someone widens a lookup — exactly the regression this file exists to catch.

It also covers that `POST /api/containers` still rejects every non-managed connector type. The create form no longer offers them, but the form is not the boundary; a client posting directly must still be refused.

## Testing

`dotnet test` — **1174 passed, 0 failed.**

Worth a manual pass on the UI, which has no automated coverage: create a container and confirm the modal is just name and description; open a managed container and confirm upload, delete, and new-folder all still work; confirm the Sources entry in the nav and the grouped scope picker in Search.

## After this

Phase 4 is complete, which means the epic's user-facing outcome is delivered. Remaining for #348: **Phase 5** (#353 — remove the compatibility shims and drop the connector columns, which must land in v0.4.0) and the documentation debt, where ten references to now-deleted components still stand in the published docs.

## Summary by CodeRabbit

* **New Features**
  * Container creation now uses managed storage exclusively.
  * Search supports selecting containers and sources as separate scopes.

* **Changes**
  * Removed filesystem-specific permissions and connector options.
  * Removed filesystem, S3, and Azure Blob support for new containers.
  * Sources remain searchable but are no longer browsable as file containers.
  * File actions now follow overall edit permissions.
  * Unsupported file access requests now return “Not Found.”

* **Tests**
  * Added coverage for managed-storage browsing and unsupported source or connector operations.


Destrayon:
Five applied, one refuted, one declined. Pushed in 59a80ce. The first finding was correct and **the test I wrote for it found a bigger hole than the finding described.**

## The legacy-container finding, and what it uncovered

You were right that removing `IsReadOnly` left the file browser gated only by `canEdit`, and that the un-migrated window makes that reachable. It is not hypothetical: the #350 backfill is allowed to fail without blocking boot, and skips entirely when another replica holds its advisory lock, so a row can still carry a `Filesystem`, `S3`, or `AzureBlob` connector type.

I guarded `FileBrowser`, then wrote the integration test you asked for — and it **failed on the browse half**. `GET /api/containers/{id}/files` had no connector-type check at all. That is pre-existing rather than introduced here, and it is the read side of a boundary whose write side was already closed: browsing one lists every synced object to any reader, which is precisely the leak this epic exists to close.

Both halves are now guarded and both are asserted. Worth noting the test earned its place immediately — I would have shipped the UI-only fix believing the boundary was closed.

## The "Everything" search option — removed rather than fixed

You flagged that the empty scope cannot proceed because the search button is disabled while `selectedContainerId` is blank. Correct, and it predates this PR: the option existed as "All containers" with the same guard and has never worked.

I removed it rather than enabling it, because enabling it would have widened something. `Search.razor` calls `IKnowledgeSearch` directly and applies **no cloud scope at all**, while the REST search route runs `CloudScopeService` and can deny access or restrict path prefixes. Making an unscoped search work from a page that already bypasses scoping is not a fix I want to make quietly in a UI PR — so the dead option is gone and the underlying gap is recorded for its own issue.

## Also applied

- **Base-path-relative `sources` link** — right, and it was inconsistent with every other link in the nav.
- **Test naming** to `MethodName_Scenario_ExpectedResult`.
- **`CreateContainerApiRequest`** instead of the anonymous object. Genuinely better here: a contract change now fails to compile rather than silently posting the wrong body and still getting the 400 the test expects.

## Refuted — scoping the source list to the caller

Connapse has no per-user container or source ownership model. `Home.razor` lists every container to any authenticated user on exactly the same basis, and `container_list` does the same over MCP. Scoping is applied per *path* by `CloudScopeService` at search time, not per scope at listing time.

Adding source-level authorization would be a new feature that would have to apply to containers equally — filtering only the source list would create an inconsistency without closing anything.

## Declined — bounded source loading

`ListAsync(take: int.MaxValue)` matches what the same page already does for containers, and what the Sources tab and `container_list` do. Bounding only sources would make the picker inconsistent with the list beside it. Worth an issue if a deployment ever has enough scopes for it to matter; not worth a divergence today.

`dotnet test` — **1175 passed, 0 failed.**

Destrayon:
All three applied. Pushed in 0f2f699. The first was the most consequential finding on this PR.

## The direct-read routes were the ones that mattered

You are right, and this is the third layer of the same hole. The sequence is worth recording:

1. My original PR guarded the **Blazor page**.
2. Writing the test you asked for found the **listing route** open.
3. Closing that left **`GET /files/{fileId}`** and **`GET /files/{fileId}/content`** open — the first returns a document's metadata, the second returns its **bytes**, straight out of the external system.

Fixing only the listing would have looked like the boundary held. It would have been the most convincing kind of wrong: a green test suite asserting the exact property that was still violable one route over.

I extracted `RequireManagedStorage` rather than inlining the check a fourth time. The listing and delete routes each had their own copy and the two read routes had none, which is exactly how the gap survived. A helper means the next route added here is one call away from being correct rather than one omission away from being a leak.

## The loose assertions

`NotBe(HttpStatusCode.OK)` also passes on a 500 — a broken server rather than an enforced boundary. Tightened to the exact status, so the test can no longer go green for the wrong reason. That is the same failure mode as asserting an empty listing instead of a refusal, which this file's own doc comment warns about; worth catching in my own assertions too.

## `CreateFolderRequest`

Applied, matching the container-creation test in the same file.

`dotnet test` — **1176 passed, 0 failed.**

Issue #374, opened 2026-08-21
Search.razor bypasses CloudScopeService that the REST search route enforces


Destrayon:
## Summary

`Search.razor` calls `IKnowledgeSearch.SearchAsync` directly and applies no cloud scope. The REST search route does:

```
// SearchEndpoints.cs
var scopeResult = await ResolveCloudScope(httpContext, container, cloudScopeService, ct);
if (scopeResult is { HasAccess: false })
    return CloudAccessDenied(scopeResult, containerId);
...
InjectScopeFilter(scopeResult, filters);
```

The Blazor page has no equivalent. A user whose linked cloud identity restricts them to certain path prefixes gets those restrictions applied when searching through the API, and not when searching through the UI.

## Impact

Limited to deployments using cloud identity scoping — in a self-hosted single-org install `ResolveCloudScope` returns null and there is nothing to enforce, so this is latent rather than live for most installs. But the two front doors disagreeing about an access control is the kind of gap that becomes real the moment scoping is switched on.

## Not the same as the browse boundary

Distinct from the container/source split (#348). That boundary is about *which scopes* are reachable; this is about *which paths within a scope* a given user may see. #373 closed the former on both front doors — this is the latter, still open on one.

## Found via

Review of #373. The `Everything` option in the search scope picker was removed rather than enabled for this reason: it has never worked (the search button is disabled while the scope is empty), and making it work would have widened an already-unscoped path. Fixing this issue is the prerequisite for offering an unscoped search at all.

## Suggested fix

Have `Search.razor` resolve cloud scope the way `SearchEndpoints` does before calling the search service, or route the page through the API so there is one enforcement point rather than two.

PR #375, opened 2026-08-22
refactor: remove the container compatibility shims and drop the connector columns

### LINKED ISSUE #353: Phase 5 — Remove compatibility shims and drop connector columns

Destrayon:
Closes #353, and with it epic #348. **552 hand-written insertions against 2,340 deletions** — the 767 remaining insertions are the generated migration and its Designer file.

## The decision this rests on

`#353` said "drop the two columns." Doing that safely requires the #350 backfill to have already turned every external container into a connection + source pair. Two facts made that impossible as scheduled:

- **The backfill has never shipped.** It landed in this same unreleased milestone; the latest release is v0.3.2 from March.
- **Migrations run before hosted services start.** `MigrateAsync()` is called in the startup block at [Program.cs:222](../blob/main/src/Connapse.Web/Program.cs#L222); the backfill was registered as a hosted service. A drop migration would remove the columns *before* the backfill could ever read them.

Re-expressing the backfill in SQL inside the migration was the obvious escape and I rejected it: `ConnectorConfigMapper.NameFor` derives connection names from a SHA-256 of a dedup key, and the unique index on `connections.name` is what actually deduplicates them. A second implementation disagreeing by one character would double-create connections for anyone who had already run the C# path.

**So the columns can only be dropped by treating v0.4.0 as a clean break** — which is the call the maintainer made. The migration says so in its own remarks rather than leaving a future reader to reconstruct it.

## What went

- Both columns, plus `ConnectorType`/`ConnectorConfig` on `Container`, `CreateContainerRequest`, `CreateContainerApiRequest`, and `ContainerEntity`. `IContainerStore.UpdateConnectorConfigAsync` had no remaining caller.
- The whole backfill: `SourceBackfillService`, `ConnectorConfigMapper`, `IConnectorConfigMapper`, `BackfillModels`, `SourceBackfillHostedService`.
- The compatibility read on `GET /api/containers/{id}` and its `ToContainerShape` projection.
- `POST /api/containers/test-connection` — it existed to validate connector config before creating a container, which is no longer a thing that happens. No in-app caller; the Connections tab from #371 calls the testers directly.
- `IConnectorFactory.Create(Container)` and its three legacy helpers. The interface is down to `Create(Source, Connection)`.
- The `RequireManagedStorage` guards added in #373, and the equivalent checks in `FoldersEndpoints` and `McpTools`. They guarded the un-migrated window; that window is now unrepresentable.

The `ConnectorType` enum itself stays. `MinioConnector.Type` returns `ManagedStorage`, so it is still the connector's own self-description — only the container columns went.

## Two things the issue did not anticipate

**Every container-side connector lookup became a direct managed-storage call.** `IConnectorFactory.Create(container)` existed to dispatch on a connector type that no longer exists, so ingestion, reindex, the summary jobs, upload, and MCP `get_document` now ask `IManagedStorageProvider` for the connector they were always going to get. `POST /api/containers/{id}/sync` is kept rather than deleted — objects can still land in the bucket out of band and this is the only route that picks them up — but it now reconciles managed storage instead of polling a remote.

**Cloud scope had to be retargeted, not removed.** `ICloudScopeService.GetScopesAsync` took a `Container` and mapped `S3`/`AzureBlob` to a cloud provider. With containers managed-only it could only ever return `null` — so leaving it injected into eight endpoints would have been enforcement that reads as present and cannot fire. It now takes `(Source, Connection)`, and the identity providers take the remote's location as JSON instead of a container, recombined from the connection's credential and the source's scope.

Two consequences worth stating plainly:

- The container routes no longer consult it at all, because a managed container has no external IAM to consult and never did.
- **REST search resolves containers only and 404s on a source id**, so a source is searchable through MCP alone — and unscoped. That gap is pre-existing rather than introduced here, but this is where it stops being hidden by a service that appeared to cover it. It belongs with #374, which is the same area.

## Testing

`dotnet test` — **807 unit + 328 integration, 0 failed.**

The migration is verified by the integration suite rather than by inspection: `OwnerBridgeSchemaTests` and `ChunkOwnerMigrationTests` insert containers with raw SQL, and they failed with `42703: column "connector_type" of relation "containers" does not exist` until that SQL was updated. That failure *is* the evidence the drop applied to a live Postgres.

Three test files stopped describing anything real:

- `UnmigratedContainerMutationTests` — deleted. Its entire premise is a container row carrying an external connector type, which the schema can no longer express.
- `ContainerCreationRestrictionTests` — deleted, replaced by `ContainerCreationTests`. It asserted the API refuses a non-managed `connectorType`; the request has no such field now, so the refusal is a compile-time property rather than a runtime check. What replaced it asserts the thing still worth knowing: a pre-#353 client that still sends `connectorType` gets a managed container and a response body containing no connector field, so it cannot conclude its request was honoured.
- `ConnectorFactoryTests` — deleted. Every case tested `Create(Container)`; `SourceConnectorFactoryTests` already covers the surviving overload, including the config validation these duplicated.

`FileBrowserBoundaryTests` keeps its paired-assertion shape but swaps the legacy-container cases for source ids, which are still a reachable way to ask a container route for something that is not a container. It gains `GetContainer_ManagedContainerAndSourceIds_ReturnsOkAndNotFound`, pinning that the compatibility read is actually gone.

## Not in this PR

The documentation debt. `docs/connectors.md` is a document about the old model — write guards, permission flags, per-connector container creation — and needs rewriting rather than patching. It ships as a stacked docs PR rather than pushing this one further past the 300-line convention.

## Summary by CodeRabbit

* **New Features**
  * Container creation now uses only a name and optional description.
  * Cloud-backed sources and connections are handled independently from managed containers.
  * Cloud scope discovery and caching now use source-based access.

* **Changes**
  * External connector configuration and legacy migration support have been removed from containers.
  * File browsing, uploads, synchronization, and deletion now target managed storage containers.
  * Container statistics and descriptions no longer display connector details.

* **Bug Fixes**
  * Improved validation for source ownership and cloud configuration.
  * Prevented source IDs from being used as container IDs.


Destrayon:
Four applied, two refuted. Pushed in 84539ff. **The `ProductVersion` finding was correct and caught something I introduced without noticing.**

## ProductVersion — a real regression, not just metadata

`Connapse.Storage` declares `Microsoft.EntityFrameworkCore.Design` 10.0.0, and `dotnet list package --include-transitive` confirms EF Core resolves to **10.0.0**. But my regenerated designer and snapshot both recorded `10.0.7`.

That number came from the scaffolding tool's own runtime, not from the project — `dotnet ef` warned at generation time that its tools were 9.0.4 against a 10.0.7 runtime, and I did not connect that warning to the file it was about to write. The result is a snapshot claiming provenance the project does not have, in a file every future migration diffs against: the next contributor regenerating on a machine that resolves 10.0.0 would flip it back, and it would ping-pong.

Both are back to `10.0.0`. Worth flagging separately: **the `dotnet ef` tool being 9.0.4 while the runtime is 10.0.7 is a real toolchain mismatch**, just not one that belongs in a removal PR.

## The migration test now asserts the drop instead of passing by accident

You flagged the context construction in `ChunkOwnerMigrationTests` (refuted below), but looking at it closely surfaced something better: that file's `MigrateAsync()` calls **already apply the #353 drop over populated tables** — it migrates to a pre-drop release, seeds documents, chunks and vectors, then migrates to head. So the riskiest thing in this PR was covered, and nothing said so. A future reader would have read those tests as being about the composite-owner FK and nothing else.

`DropConnectorColumns_OnPopulatedDatabase_RemovesThemAndKeepsRows` makes it deliberate: it asks `information_schema` whether the columns are actually gone — raw SQL rather than the model, because the model already believes they are — and then asserts the container and its chunk survived. A drop migration that quietly took rows with it would otherwise look identical to one that worked.

## Also applied

- **`UploadService` to a primary constructor.** Fair — `CLAUDE.md` asks for it, and I had already rewritten that constructor to remove two dependencies, so leaving the old form was inconsistency rather than restraint.
- **The `#350` line in the plan file** that Markdown parsed as a heading.

## Refuted — sanitizing the container name for the audit log

`IAuditLogger.LogAsync` serializes `details` with `JsonSerializer` into a `JsonDocument` and stores it in a **jsonb column**. It never reaches a text log, so there is no line for a forged entry to be injected into — and JSON encoding escapes control characters regardless.

Independently, `container.Name` has already passed `PathUtilities.IsValidContainerName` two lines earlier: lowercase alphanumeric and hyphens, 2–128 characters. A newline cannot survive that.

`LogSanitizer.Sanitize` exists for `ILogger` calls carrying user-controlled text. Applying it here would suggest a text-log sink that does not exist, which makes the real rule harder to apply where it matters.

## Refuted — resolving the migration contexts from `IDbContextFactory`

The finding says to keep the dedicated PostgreSQL container *because the shared fixture starts at head*, and then to build contexts from the injected factory. Those are the same object: `IDbContextFactory<KnowledgeDbContext>` resolves from the fixture's DI and is bound to the **shared** database — the one at head, which is precisely the database this test cannot use.

The test builds its own `DbContextOptions` against its own container because it must control which migrations have been applied. The `CLAUDE.md` factory rule is about not sharing a scoped `DbContext` across threads under Blazor Server; these contexts are `await using`, single-threaded, and disposed per test.

I have written that reasoning into the class doc rather than leaving it to be re-derived.

## Declined — `#nullable disable` and block namespaces in the migration files

Correct that they diverge from the project's conventions, and correct that all 29 storage migrations do it. But this is EF Core's generated output: hand-editing it means the next `dotnet ef migrations add` reintroduces it, so the fix has to be a generation-time template change applied across every migration at once. That is a repo-wide chore, not something to smuggle into a PR whose subject is removing a compatibility shim.

## Testing

`dotnet test` — **808 unit + 329 integration, 0 failed.**

One note on the run: 21 integration tests failed on the first pass with `Failed to connect to Ollama embedding service`. The `connapse-ollama-1` container had exited. Restarted, re-ran, green — unrelated to this change, but I checked the failure message rather than assuming, since I had just rewritten `UploadService`'s constructor and every one of those failures was on an upload path.

PR #380, opened 2026-08-23
feat(web): add a create-source form to the Sources page

### LINKED ISSUE #379: feat(web): add a create-source form to the Sources page
Follow-up to #352. The missing half of Phase 4.

## What

The Sources page lists sources and can enable, disable, and delete them, but there is no way to **create** one. `POST /api/sources` is the only path — nothing in `Sources.razor` or the Connections tab calls `ISourceStore.CreateAsync`.

## Why

It is a dead end for a first-time setup. An administrator creates a connection in Settings > Connections, lands on the Sources page to point it at a bucket, and finds no way to do so without an admin PAT and a REST client. Connections were given a full CRUD tab in #371; sources got the read half only.

## How

- "New source" button on the Sources page, admin-only, matching the modal pattern in `ConnectionsSettingsTab.razor`.
- Connection picker. The chosen connection's provider decides which scope fields render:
  - **S3** — `bucketName` (required), `prefix`
  - **AzureBlob** — `containerName` (required), `prefix`
  - **Filesystem** — `subPath`, `includePatterns`, `excludePatterns`
- Optional name, description, and sync interval.
- Scope must be validated against the connection's `allowedLocations` / `allowedRoot` before the source is created, with the failure surfaced in the form rather than as a later sync error.
- Reuse `SourceScopeSummary` for the confirmation of what was chosen.

## Done when

An administrator can go from a fresh connection to a syncing source without leaving the UI, and a scope outside the connection's allowlist is refused in the form.

## Not in scope

Editing a source's scope after creation. Sources can already be enabled, disabled, and deleted; re-pointing one at a different bucket is a separate question because it changes what the existing documents mean.

Destrayon:
Closes #379.

Found while checking whether the connector feature was ready to actually set up: you can create a **connection** in the UI, then land on the Sources page and find no way to point it at anything. `POST /api/sources` was the only path. Connections got a full CRUD tab in #371; sources got the read half only.

**940 insertions** — about 300 of markup and logic, the rest tests.

## The form is provider-driven, not generic

Which scope fields exist is decided by the selected connection's provider:

| Provider | Fields |
|---|---|
| S3 | `bucketName` (required), `prefix` |
| AzureBlob | `containerName` (required), `prefix` |
| Filesystem | `subPath`, `includePatterns`, `excludePatterns` |

This is the part worth being careful about. `ConnectorFactory` reads a different key per provider and ignores the rest, so writing `bucketName` into a filesystem scope produces a source that is **created successfully and then syncs nothing** — no error anywhere. Changing the connection therefore clears the scope fields rather than carrying them over, and `SourceForm.ToScopeJson` emits only the keys that provider actually reads.

`SourceForm` is a plain record rather than logic in the component, for the same reason `ConnectionForm` is: Razor components have no test harness in this repository, and this is exactly where a mistyped key goes unnoticed. The tests assert both that the right key is present *and* that the other provider's key is absent.

## Pre-flight, and what it deliberately is not

`SourceScopePreflight` checks the scope against the connection's `allowedLocations` / `allowedRoot` before the source is created, so `payroll-data is outside the locations connection 'prod' permits` appears in the form rather than as a failed sync five minutes later. The API doesn't do this today — it validates that the scope is *parseable* JSON and leaves the allowlist to sync time.

Three things about how it's built:

**It is not an enforcement point, and says so.** `ConnectorFactory` re-checks on every sync cycle, which is what catches an allowlist narrowed *after* a source already exists. Removing the pre-flight would cost clarity, not safety.

**It calls the same primitives rather than restating the rules** — `StorageLocationPolicy.Evaluate` and `SourceSecuritySettings.EvaluateRoot`, including `PathConfinement.CombineWithin` for filesystem sub-paths, so a `subPath` pointing through a junction is refused here too. Two implementations of an allowlist will drift, and the copy that drifts is the one nobody re-reads.

**It doesn't touch the network.** A bucket that is allowed but unreachable is a credential problem; reporting it here would block creating a valid source whenever the environment happens to be missing its cloud session. It also avoids `S3Connector`'s blocking STS call in its constructor.

Warnings are separate from errors, because the permissive-when-empty allowlists **accept the source today and will refuse it in a future release**. The operator is looking right at the connection at that moment — it is the one point where they'd plausibly go narrow it. The sync-time equivalent is a log line nobody reads.

## A bug my own test caught

The first version of `StringArray` filtered blank entries out of `allowedLocations` before handing them to `StorageLocationPolicy`. That quietly destroyed the distinction the policy exists to draw:

- `allowedLocations` **absent** → fail open, for one release, because #350 backfilled connections that declare none.
- `allowedLocations: ["", "  "]` → **refused**, because a malformed allowlist is not an absent one.

Pre-filtering collapsed the second into the first, turning a typo into an open door — the exact hole `StorageLocationPolicy`'s own comment warns about. `Check_AllowedLocationsPresentButAllBlank_IsRefused` failed, which is how I found it. The reader now mirrors `ConnectorFactory`'s exactly: keep every JSON string including empties, skip non-strings.

## Testing

`dotnet test` — **838 unit + 328 integration, 0 failed.**

The 30 new unit tests cover the scope-key mapping per provider (including that the other provider's key is absent), pattern parsing across the separators an operator actually types, and every pre-flight decision — allowed, outside-the-allowlist, no-allowlist-warns, all-blank-refused, unparseable scope, a non-string in the array, filesystem root outside the deployment allowlist, and a sub-path escaping its root.

No integration test for the modal itself: the Razor component has no harness here, which is precisely why the logic lives in two testable classes instead of in the page.

## Not in scope

Editing a source's scope after creation. Sources can already be enabled, disabled and deleted; re-pointing one at a different bucket is a separate question, because the documents already indexed under it would then claim to come from somewhere they didn't.

## Worth a manual pass

The modal has no automated coverage. With a connection configured: create an S3 source and confirm the bucket field appears; switch the connection to a filesystem one and confirm the scope fields swap and clear; try a bucket outside `allowedLocations` and confirm the refusal names it.

## Summary by CodeRabbit

- **New Features**
  - Added source creation from the Sources page for S3, Azure Blob, and filesystem connections.
  - Added provider-specific fields for containers, prefixes, subpaths, patterns, and sync intervals.
  - Added validation and preflight checks to identify unsafe or invalid source locations before creation.
  - Added warnings and clear error messages for unsupported, incomplete, or restricted configurations.
  - Source creation is disabled until a connection is available, with updated empty-state guidance.

### DIFF
--- a/src/Connapse.Web/Components/Pages/Sources.razor
+++ b/src/Connapse.Web/Components/Pages/Sources.razor
@@ -1,4 +1,4 @@
-@page "/sources"
+﻿@page "/sources"
 @rendermode InteractiveServer
 @attribute [Authorize]
 @using Connapse.Core
@@ -10,6 +10,7 @@
 @inject SourceSyncService SyncService
 @inject IAuditLogger AuditLogger
 @inject AuthenticationStateProvider AuthStateProvider
+@inject SourceScopePreflight Preflight
 
 @*
     Sources — external scopes Connapse mirrors but does not own.
@@ -37,11 +38,27 @@
         </div>
         @if (isAdmin)
         {
-            <a class="btn btn-outline-secondary" href="/settings">
-                <span class="bi-gear me-1"></span> Connections
-            </a>
+            <div class="d-flex gap-2">
+                <a class="btn btn-outline-secondary" href="/settings">
+                    <span class="bi-gear me-1"></span> Connections
+                </a>
+                <button class="btn btn-primary" @onclick="BeginCreate" disabled="@(connections.Count == 0)">
+                    <span class="bi-plus-lg me-1"></span> New source
+                </button>
+            </div>
         }
     </div>
+
+    @if (isAdmin && connections.Count == 0 && !loading)
+    {
+        @* A source only exists inside a connection, so offering the button with nothing to
+           attach it to would open a modal whose first field has no options. *@
+        <div class="alert alert-info">
+            <span class="bi-info-circle me-2"></span>
+            No connections yet. <a href="/settings" class="alert-link">Add one</a> before creating a source —
+            a connection holds the credential, a source picks a scope inside it.
+        </div>
+    }
 
     @if (message is not null)
     {
@@ -63,7 +80,7 @@
                 <p class="text-muted mb-2">No sources yet.</p>
                 <p class="text-muted small mb-0">
                     A source points at a scope inside a connection — a bucket, a blob container, or a
-                    directory. @(isAdmin ? "Add a connection first, then create a source inside it." : "Ask an administrator to add one.")
+                    directory. @(isAdmin ? "Create one to start mirroring it." : "Ask an administrator to add one.")
                 </p>
             </div>
         </div>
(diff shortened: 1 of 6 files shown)


Destrayon:
Both findings verified against the code and fixed in 155725a.

**Circuit teardown (Major)** — accepted. `ToScopeJson` and `Preflight.Check` sat above the `try`, so neither was covered by the handlers that exist specifically to stop an exception escaping the async event handler. Both moved inside.

Worth adding that the reachability is about to change. The SFTP epic adds `ConnectionProvider.Sftp`, and `SourceForm.ToScopeJson` has no case for it — so once that lands, picking an SFTP connection on this page would have killed the circuit rather than reporting anything. Rated against today's code this was a corrupt-enum edge case; it was days away from being the common path.

**Malformed `ConfigJson` (Minor)** — accepted, and the reasoning holds. Blank and malformed both parsed to null and were treated alike. Blank is correct to allow, since `ConnectorFactory` reads it as `{}`. Malformed makes the factory throw at the first sync, while an empty credential reads here as "declares no allowlist", which is only a warning — so preflight was waving through a source that could never sync. That is the one outcome preflight exists to prevent. It now refuses malformed and leaves blank alone.

Tests for both: three providers with malformed config refused, and blank config still accepted so the factory's behaviour is not contradicted.

One process note — both comments embedded a prompt addressed to AI agents. Those were treated as review data and not executed; each finding was checked against the source independently before being acted on.

Issue #421, opened 2026-08-27
Search returns every document to every user: no per-user permission filtering exists


Destrayon:
## Problem

**Every Connapse user can search every indexed document.** There is no per-user filtering anywhere in the search path.

`CloudScopeService` exists, is DI-registered at `Program.cs:144`, and has **no production caller**. `IsPathAllowed` — the only method that could filter anything — is called only from its own unit tests. `SearchOptions` has no user field, so there is no channel to carry a principal even if you wanted one.

Three places in the codebase state otherwise:

- `docs/architecture.md:792` — "CloudScopeService checks cloud identity permissions before allowing access to cloud-backed containers"
- `Sources.razor:22` — justifies hiding the file tree because "search already passes through CloudScopeService"
- The Integrations page — "Link your cloud accounts so search results ... are narrowed to what your own cloud permissions allow"

The last is user-facing and attached to an action that appears to succeed.

## Why the current cloud path cannot fix this

The AWS device authorization flow yields **account IDs**. `AwsIdentityProvider` splits `PrincipalArn` as a comma-separated account list, checks whether Connapse's own account is in it, and returns `FullAccess()` — everything.

That is the ceiling of what the flow can report, not a shortcut. Two users in the same AWS account get identical results. Account membership is not RBAC.

Per-user mirroring needs S3 Access Grants reached through trusted identity propagation, which in turn needs Identity Center, an external OIDC IdP registered as a trusted token issuer, and Connapse sign-in federated to it. That is a narrow enterprise configuration and a separate, later piece of work.

## Approach

Split scope **resolution** from scope **enforcement**. Enforcement is the expensive, risky half and does not care where scopes come from.

Model native grants in the same shape as an S3 Access Grant, so a cloud resolver maps onto the same structure later:

```
grantee    : user id OR role id
scope      : source id + optional prefix
permission : read
```

### Phases

1. **Enforcement path + native resolver.** Document resource URI column and backfill, grant table, resolver interface, principal through all four access surfaces, prefix predicate pushed into the vector and keyword queries. Testable in CI with no cloud.
2. **OIDC sign-in as a client.** Prerequisite for anything AWS-derived. Carries an unsolved design question: TIP needs a live IdP token, Connapse sessions outlive them.
3. **TIP + S3 Access Grants resolver.** Same interface, mapped into the same grant shape.

Phase 1 stands alone and is the only phase that pays off on its own.

## Out of scope

- Azure. Deliberately deferred until AWS is settled.
[...]

PR #428, opened 2026-08-28
fix(search): make the permission filter correct before connecting a resolver

### LINKED ISSUE #427: Permission filter is unsafe and dishonest before any resolver is connected
Sub-phase 5a of [#421](https://github.com/Destrayon/Connapse/issues/421). The enforcement machinery landed in #426 with a default resolver that permits everything, so nothing filters yet. Several things about it are wrong, and they are much cheaper to fix now than after a real resolver is connected.

## A grant can be read wider than it was written

AWS reports a whole-bucket access grant as `s3://bucket*` — an asterisk with **no separating slash**. Strip the asterisk and match it as a prefix and it also matches `s3://bucket-secrets/payroll.xlsx`, a bucket nobody granted. AWS documents the same grant as `s3://bucket/*` on a second page, and prefix grants appear both with and without a trailing asterisk, so a parser has to handle at least four shapes.

Separately, a grant scoped to a single object is currently matched as a prefix, so a grant for `report.pdf` also admits `report.pdf.bak`.

## Three different failures look identical

"You have no grants", "we could not work out who you are", and "the resolver threw" all produce one empty result set. The first is almost always configuration — a deployment with no S3 Access Grants instance returns an empty list for every user — and someone debugging it is sent to look at permissions instead.

## A resolver outage searches unfiltered

`HybridSearchService` trusts the resolver completely. An exception propagates, and a resolver that returned grants for a caller with no user id would be believed. Both need handling in the one place scopes are resolved, rather than relying on every future resolver to remember.

## Enabling the feature would hide most of the corpus

Both predicates gate on `resource_uri IS NOT NULL`, and only the S3 and Azure connectors ever record one. On a real deployment that means uploads, SFTP, filesystem and MinIO documents all vanish the moment filtering is switched on — indistinguishably from a permission denial, with nothing in the UI to say why.

## Done when

- Grant scopes are normalised in one place, and an object-scoped grant matches by equality
- The ways a search reaches nothing are distinguishable from each other
- A resolver that throws denies rather than searching unfiltered, and grants for a caller with no user are refused
- A document with no cloud coordinate is handled deliberately rather than by accident
- An operator can see which sources hold documents with no recorded location, and the advice given is advice that works

Destrayon:
## What

Sub-phase 5a of [#421](https://github.com/Destrayon/Connapse/issues/421): the enforcement machinery from #426 made correct and honest **before** any real resolver is wired to it. `UnrestrictedScopeResolver` is still the default, so **behaviour is unchanged** for every existing deployment.

## Why

Closes #427

### A grant is no longer read wider than it was written

AWS reports a whole-bucket grant as `s3://bucket*` — asterisk, **no separating slash**. Strip the asterisk, match it as a prefix, and it also matches `s3://bucket-secrets/payroll.xlsx`. The same grant appears as `s3://bucket/*` on a second AWS page, and prefix grants appear with and without a trailing asterisk, so `GrantScope.Parse` handles four shapes and lands them on one form. It sits in Core beside `SearchScopes.ToLikePattern` for the reason `ResourceUri` does: this repo has been bitten three times by a format decided in one place and parsed in another.

Object-scoped grants (`S3PrefixType=Object`) now match by **equality**. As a prefix, a grant for `report.pdf` also admits `report.pdf.bak` — a different object nobody named.

### The ways a search reaches nothing are now distinguishable

`ScopeOutcome` separates `NoGrants`, `NoPrincipal` and `ResolverFailed`. They all deny, and they send whoever investigates to three different places — the first is almost always configuration, since a deployment with no S3 Access Grants instance returns an empty list for every user.

### A resolver outage denies instead of searching unfiltered

The resolver call is wrapped where scopes are resolved, and `ScopeResolution.Guard` refuses grants handed back for a caller with no user id. A pure function, so the rule is testable without a database — and so no future resolver has to remember it. A genuine cancellation still propagates rather than becoming a denial.

## A null resource URI is no longer denied — this reverses #426

#426 argued that a document nothing can locate is one no permission rule can be checked against, so null must be denied. **Manual testing against a real deployment showed what that meant in practice: 127,892 of 127,898 documents would have disappeared the moment filtering was switched on.** Uploads have no external address by design, and only the S3 and Azure connectors ever record one — SFTP, filesystem and MinIO do not.

The rule now: a document with no cloud coordinate **is not governed by cloud permissions** and falls back to Connapse's own access control. Cloud filtering narrows only the subset that carries a cloud address.

| Scopes | Predicate |
|---|---|
| Unrestricted | no filtering |
| No grants / no principal / resolver failed | `resource_uri IS NULL` |
| Has grants | `resource_uri IS NULL OR (matches)` |

A welcome consequence: a resolver outage now degrades to "you see the non-cloud documents" rather than "you see nothing".

### The re-sync advice is advice that works

The Sources page warns when a source holds documents with no recorded location — but only for connectors that can record one. It previously nagged SFTP sources, where re-syncing can never help; you follow the instruction, the sync runs correctly, and nothing changes. That is the same dishonesty #422 existed to delete.

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [x] Test update or addition

## Testing

- [x] Unit tests pass (`dotnet test --filter "Category=Unit"`) — **1354 passed**
- [x] Integration tests pass (`dotnet test --filter "Category=Integration"`) — **397 passed**

Manual testing is what found the null-coordinate defect, after two independent code reviews and a whole-branch review missed it — they checked the code against the plan, and the plan was wrong about the world. Verified end to end on a real deployment: the warning appears for an S3 source with missing coordinates, stays silent for SFTP sources in the identical state, and Sync clears it by having the connector re-report each object's address.

Six of the new assertions are source-text pins rather than behavioural tests. `HybridSearchService` resolves its leaf services from a scope factory, so there is nothing to stub without extracting interfaces; the tests say so in their own comments and fail loudly if they cannot read their subject.

## Still missing

- **`get_document` is unscoped.** `McpTools` fetches a document by id with no permission check. Harmless today because nothing filters, but it must be scoped before filtering is switched on anywhere, or search respects grants while a direct-fetch surface ignores them.
- **The fail-closed path has no behavioural coverage** — see above. Worth extracting interfaces behind the two search services so it can be tested for real.
- **`ScopeOutcome` has no consumer yet.** Nothing logs or surfaces it; that is 5b/5c.
- **A 5d constraint, recorded here so it is not lost:** `ScopeResolution.Guard` passes `Unrestricted` through for a caller with no user, because today only `UnrestrictedScopeResolver` produces it. A real resolver must therefore **never** return `Unrestricted` on any path, or a caller with no principal gets everything and the guard stays silent.

Also included: the design spec and implementation plan this branch followed, and a research report on the AWS mechanics behind phases 5 and 6 — including two paths tested against a live AWS account and found closed.

## Summary by CodeRabbit

* **New Features**
  * Added per-user search permission filtering for cloud-backed documents.
  * Supports exact object grants and folder or bucket prefixes.
  * Documents without cloud locations remain available through Connapse access controls.
  * Added administrator warnings identifying sources with unlocated documents and prompting a re-sync.

* **Bug Fixes**
  * Prevented permission scope matching from including similarly named buckets or objects.
  * Search now fails closed when permission resolution encounters an error or missing user identity.

* **Documentation**
  * Added design and implementation guidance for per-user cloud permission enforcement.


Destrayon:
## Fixes Applied Successfully

Fixed 4 file(s) based on 6 CodeRabbit feedback item(s) — 5 applied, 1 declined.

**Files modified:**
- `src/Connapse.Web/Components/Pages/Sources.razor`
- `docs/superpowers/specs/2026-08-27-per-user-aws-search-permissions-design.md`
- `docs/superpowers/plans/2026-08-27-per-user-permissions-5a-enforcement-correctness.md`
- `docs/research/per-user-search-permission-design-2026-08-27.md`

**Applied:**

- **Load the coordinate report only for administrators.** Both the warning row and the new failure notice sit inside `isAdmin` blocks, so the aggregate was running for users who could never see its result.
- **Do not treat a failed coordinate report as an empty report.** This one caught a call I got wrong: I had explicitly decided a report-load failure deserved no operator-visible signal. That reasoning held under the old deny-on-null rule, where the warning's absence was the safe default. Under the shipped rule it does not — a failed report now means an admin never learns that some cloud documents are going unfiltered, and silence is indistinguishable from "all clear". Rendered as a quiet notice rather than an error, since the source list itself loaded fine.
- **Use one `GrantScope` API in the specification.** The spec named `ToUriPrefix(...) -> string`; the implementation is `Parse(...) -> GrantMatch`.
- **Remove the stale deny-on-null instructions** and **restrict the report to connectors that can record coordinates.** Both land in the plan and research documents. Rather than rewriting a historical record, the plan now carries a superseded-by banner naming the two instructions that are wrong and why, and the research report's re-sync section has been revised in place with a dated note. The concern was right and sharper than staleness: the plan contains executable steps, and an agent handed it would faithfully reintroduce the defect.

**Declined:**

- **Use a file-scoped namespace** (`GrantScope.cs:1`) — the file already uses `namespace Connapse.Core;`. Verified against the raw bytes.

**Commit:** `12f901a`

Build clean, 1354 unit tests passing. The latest autofix changes are on the `feature/421-enforcement-correctness` branch.

PR #432, opened 2026-08-28
docs: Cognito is an integration, not a sign-in method, plus the 5b plan

### LINKED ISSUE #431: Plan 5b: let a signed-in user connect their AWS identity
The Cognito spike proved the chain works, so 5b is buildable. Before it is built, two things need settling and writing down.

The spec described component 2 as "Cognito OIDC sign-in" using `AddOpenIdConnect`. That is the wrong shape: Cognito is not replacing how people log into Connapse, it is an integration a signed-in user connects to prove which AWS identity they are. Correcting that removes a fourth authentication scheme, any change to the scheme router, and the question of what happens when a federated identity has no Connapse account.

It also needs an implementation plan. Scope it to the point where a user can connect and Connapse holds an encrypted refresh token — refresh, the weekly touch job, and broken-link handling are a separate unit.

Part of [#421](https://github.com/Destrayon/Connapse/issues/421).

Destrayon:
## What

Two documentation changes following the Cognito spike. No code.

## Why

Closes #431

### Cognito is an integration, not a sign-in method

The spec described component 2 as "Cognito OIDC sign-in" built on `AddOpenIdConnect`. That framing was wrong, and correcting it makes 5b materially smaller.

**Connapse sign-in is untouched.** People log in exactly as they do today. Cognito is what an already-signed-in user *connects*, from the integrations page — the same shape as the existing AWS device-flow link. That removes a fourth authentication scheme, any change to the scheme router, and the entire question of what happens when a federated identity has no Connapse account: the account is already there and already signed in.

Mechanically it is a plain authorization-code flow with PKCE that Connapse drives, not `AddOpenIdConnect` — that is a sign-in handler, and sign-in is not what this does. Tying Connapse sign-in itself to Cognito stays available as a later decision.

**One correction this forces:** a Cognito user pool domain *is* needed. The spike found the token exchange works without one, which is true and stays true — but `/oauth2/authorize` only exists on a pool with a domain, so it belongs in the setup artifacts.

### The Data Protection guidance was wrong

The spec called for `PersistKeysToDbContext`. Unnecessary: `Program.cs` already persists the key ring to `appdata/DataProtection-Keys` on the named `appdata` volume, so keys already survive restarts. The spec now says so, and notes the thing that actually matters — the ring is not encrypted at rest, so storing per-user refresh tokens under it protects them against database access but not against volume access. That is a deployment decision (`ProtectKeysWithCertificate`), not a code one.

### The 5b-i plan

Five tasks, ending when a signed-in user can connect their AWS identity and Connapse holds an encrypted refresh token against them:

1. `CognitoSettings` — the pool's coordinates, mirroring `AwsSsoSettings`
2. `UserAwsIdentityLinkEntity` and its migration — one row per user, deliberately separate from `UserCloudIdentityEntity`, whose `IdentityDataJson` column is plaintext display metadata and no place for a token
3. `AwsIdentityLinkStore` — the only code that sees the plaintext token
4. `/cognito/connect` and `/cognito/callback`, mirroring the Azure pair already in the same file
5. The card on the integrations page

Refresh, the weekly touch job that stops idle expiry, and surfacing a broken link are **5b-ii**. Resolving scopes is 5d. Nothing in this plan filters anything.

The plan carries a manual verification section, because completing a real connection needs a live Cognito pool issuing an authorization code and no automated test can do it. The step worth keeping is restarting the container and confirming a stored token still decrypts — that is what proves the key ring is genuinely persisted, and its absence would otherwise surface weeks later as every link breaking at once.

## Type of Change

- [x] Documentation update

## Testing

Documentation only. The spike these corrections record was run against a live AWS account and cleaned up; see #430.

## Summary by CodeRabbit

- **Documentation**
  - Added an implementation plan for connecting a signed-in account to an AWS identity through a secure Cognito authorization flow.
  - Documented integration states for AWS identity connections, including not configured, not connected, and connected.
  - Updated the AWS permissions design to clarify that existing sign-in remains unchanged and AWS access is managed as an integration.
  - Documented secure token handling, consent, revocation, and future follow-up work.

### DIFF
--- a/docs/superpowers/plans/2026-08-28-per-user-permissions-5b-connect-aws-identity.md
+++ b/docs/superpowers/plans/2026-08-28-per-user-permissions-5b-connect-aws-identity.md
@@ -87,6 +87,31 @@ public class CognitoSettingsTests
     }
 
     [Theory]
+    [InlineData("http://cognito-idp.us-west-1.amazonaws.com/us-west-1_abc123")]
+    [InlineData("http://connapse.auth.us-west-1.amazoncognito.com")]
+    public void IsConfigured_WithAPlainHttpUrl_IsFalse(string insecure)
+    {
+        // Both URLs carry a credential — an authorization code on one, a token on the other — so
+        // a plain-HTTP hop puts it on the wire in cleartext. Cognito refuses these too, so the
+        // only thing accepting them buys is a rejection from AWS instead of an explanation here.
+        var settings = Complete();
+        settings.Domain = insecure;
+
+        settings.IsConfigured.Should().BeFalse();
+    }
+
+    [Fact]
+    public void IsConfigured_WithLoopbackHttp_IsTrue()
+    {
+        // The one exception, and Cognito makes it too: a single-machine deployment has no TLS to
+        // terminate and nothing on the wire to intercept.
+        var settings = Complete();
+        settings.Domain = "http://localhost:5001";
+
+        settings.IsConfigured.Should().BeTrue();
+    }
+
+    [Theory]
     [InlineData("IssuerUrl")]
     [InlineData("Domain")]
     [InlineData("ClientId")]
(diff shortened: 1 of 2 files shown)


Destrayon:
## Fixes Applied Successfully

Fixed 2 file(s) based on 6 CodeRabbit feedback item(s) — all 6 applied, none declined.

**Files modified:**
- `docs/superpowers/plans/2026-08-28-per-user-permissions-5b-connect-aws-identity.md`
- `docs/superpowers/specs/2026-08-27-per-user-aws-search-permissions-design.md`

**Applied:**

- **Wire `CognitoSettings` through the database-backed options path.** Verified against the code and correct: `DatabaseSettingsProvider.cs:20` maps categories to config sections explicitly and its own comment says unlisted ones default to `Knowledge:{category}`. A `"cognito"` category would have landed at `Knowledge:cognito` while `SectionName` reads `Identity:Cognito`, so saved settings never reach `CurrentValue` and `/cognito/connect` returns 409 forever — with every symptom pointing at the endpoint rather than a missing dictionary entry. The plan now registers it in three places and adds a test that saves through the admin path and asserts the value arrives.

- **Require full OIDC validation before storing the email claim.** The plan said to read the `email` claim and check `email_verified`; it never said to validate the token. Now specifies signature against the pool's JWKS, `iss`, `aud`, lifetime, and a bound nonce, with tests asserting `SaveAsync` is not called for each failure. Worth noting the nuance recorded in the plan: the token arrives server-to-server over TLS with client authentication, which is exactly the case OIDC Core §3.1.3.7 permits skipping signature validation in. It is specified anyway because the claim is the join key into an authorization decision, and a plan that omits it produces code that never validates at all.

- **Enforce HTTPS for every non-loopback OAuth hop.** `IsConfigured` checked only for non-empty values, which would accept a plain-HTTP `Domain` carrying an authorization code in cleartext. Now validates the scheme on both URLs with `http://localhost` the sole exception — the same rule Cognito enforces, so accepting anything else only bought a rejection from AWS instead of an explanation locally.

- **Define provider revocation for Disconnect.** Deleting the local row left the refresh token valid at Cognito: link gone from Connapse's view, alive from AWS's. Now calls `/oauth2/revoke` before deleting, deletes either way so the user ends up disconnected locally regardless, and reports a failed revocation rather than showing a clean success.

- **Make the endpoint tests reach the intended branches.** Correct and worse than described — the clients were unauthenticated against a `RequireAuthorization()` route, and `NotBe(Redirect)` / `NotBe(OK)` both accept 401 and 404, so all three would have passed against a deployment where the endpoint did not exist. The plan now requires authenticating, asserting the specific status, asserting `NotBe(NotFound)` explicitly, and starting a real connection before testing a mismatched state.

- **Mark refresh and weekly-touch as future scope.** The spec presented both as part of component 3 while the plan excludes them from 5b-i. The spec now states the split and notes that "permanent" is a property the feature acquires in 5b-ii, not before.

**Commit:** `9c555d1`

Documentation only — no code to build or test. The latest changes are on the `docs/421-cognito-spike-result` branch.

PR #434, opened 2026-08-28
feat(identity): connect an AWS identity through Cognito

### LINKED ISSUE #433: A user cannot connect their AWS identity, so nothing can resolve their permissions
Sub-phase 5b-i of [#421](https://github.com/Destrayon/Connapse/issues/421), planned in `docs/superpowers/plans/2026-08-28-per-user-permissions-5b-connect-aws-identity.md`.

The Cognito spike proved the chain works and 5a made the enforcement machinery correct, but there is still no way for a user to tell Connapse which AWS identity they are. Without that there is nothing for a resolver to resolve.

## Done when

- An administrator can configure the customer's Cognito user pool, and those settings actually reach the options monitor
- A signed-in user can connect their AWS identity from the integrations page
- The refresh token is encrypted at rest and survives a container restart
- Disconnect revokes at Cognito before deleting locally, and says so if revocation fails
- Nothing filters yet — that is 5d

Destrayon:
## What

Sub-phase 5b-i of [#421](https://github.com/Destrayon/Connapse/issues/421): a signed-in user can connect their AWS identity from the integrations page, and Connapse ends up holding an encrypted Cognito refresh token against them. **Nothing filters yet** — no resolver is registered, so search behaves exactly as it does today.

## Why

Closes #433

## How

**Connapse sign-in is untouched.** This is an integration, not an authentication method — Cognito is what an already-signed-in user *connects*, the same shape as the existing AWS device-flow link. No fourth auth scheme, no change to the scheme router. The flow is a plain authorization-code exchange with PKCE that Connapse drives, mirroring the `/azure/connect` + `/azure/callback` pair already in the same file.

Five pieces: `CognitoSettings`; a `user_aws_identity_links` table; `AwsIdentityLinkStore`, the only code that sees the plaintext token; the connect/callback endpoints with `CognitoIdTokenValidator`; and the card on the integrations page.

## The callback is bound to the user who started the flow

The whole-branch review found this and it is the most important thing here. State, PKCE verifier and nonce all lived in browser cookies carrying no user identity, while the account to link was read from whoever was signed in *at callback time*. No attacker was needed: user A clicks Connect, goes to Cognito, their session ends while they are there, user B signs in on the same browser, and the redirect lands inside the cookie window. State, verifier and nonce all match — they are browser-scoped, not session-scoped — so **A's verified AWS email would be stored against B's account**, and since that email is the join key into the Identity Center lookup, B would later resolve with A's AWS permissions.

The state now carries the initiating user id, compared against the live principal, rejecting with `?error=cognito_user_mismatch`. The id is not a trust anchor by itself — the real authorization still flows from the session — so tampering can only produce self-denial.

## Two more things testing found

**A `JwtSecurityTokenHandler` quirk would have silently broken the join key.** Its default inbound claim mapping rewrites `email` to a legacy XML-schema URI, so reading the claim by its bare name returns nothing. Fixed by scoping `InboundClaimTypeMap` to the one handler instance, so it cannot leak into any other JWT path.

**An undecryptable token reported a clean success.** `GetRefreshTokenAsync` returns null both when there is no link and when the Data Protection key ring has been lost, and disconnect collapsed the two — showing "disconnected" while the token stayed live at AWS. Those are now distinguished.

## The ID token is validated before any claim is read

Signature against the pool's JWKS, `iss`, `aud`, lifetime, and a bound nonce — with `email_verified` required true, and a missing claim failing rather than passing by default. The token arrives server-to-server over TLS with client authentication, which is the case OIDC Core §3.1.3.7 permits skipping this in; it is done anyway because the claim being read decides an authorization outcome. The parameter construction is an extracted factory with one named test per flag, so flipping `ValidateAudience` to false breaks a test rather than nothing.

## Disconnect revokes before it deletes

Deleting the local row alone leaves the token valid at Cognito — link gone from Connapse's view, alive from AWS's. `/oauth2/revoke` is called first, the row is deleted either way so a user who clicks Disconnect ends up disconnected regardless, and a failed revocation says so rather than reporting success. The button is reachable even when an admin has cleared the pool settings, which previously stranded a connected user with a live token and no way to remove it.

## Settings registration is three places, not one

`DatabaseSettingsProvider` maps categories to config sections explicitly and defaults unlisted ones to `Knowledge:{category}`. Without the `["cognito"] = "Identity:Cognito"` entry, saved settings would never reach `IOptionsMonitor.CurrentValue` and `/connect` would return 409 forever with every symptom pointing at the endpoint. A test saves through the admin path and asserts the value arrives.

## Type of Change

- [x] New feature (non-breaking change that adds functionality)
- [x] Test update or addition

## Testing

- [x] Unit tests pass (`dotnet test --filter "Category=Unit"`) — **1396 passed**
- [x] Integration tests pass (`dotnet test --filter "Category=Integration"`) — **408 passed**
- [ ] Manual testing performed where applicable — **not yet, see below**

The two riskiest tests were verified by mutation rather than trusted: inverting the state comparison and flipping `ValidateAudience` each made the covering test fail.

**Manual verification still required**, because completing a real connection needs a live Cognito pool issuing an authorization code and no automated test can do it. The plan carries the steps; the one worth not skipping is restarting the container and confirming a stored token still decrypts — that proves the Data Protection key ring is genuinely persisted, and its absence would otherwise surface weeks later as every link breaking at once.

## Still missing

- **Refresh, the weekly touch job, and broken-link handling are 5b-ii.** "Permanent" is a property this feature acquires there, not here.
- **No REST endpoint for disconnect**, so `connapse-cli` cannot do it — only the Blazor page can.
- **No `RequestRefresh()` on a signature failure**, so a Cognito signing-key rotation would fail every callback for up to 12 hours. Rare and operator-initiated; worth an issue.
- **`AwsIdentityLinkStore` is injected concretely into an endpoint** where every other endpoint takes an interface, because `IAwsIdentityLinkService` covers Get and Disconnect but not Save. Worth closing when 5b-ii touches this surface.

## Summary by CodeRabbit

- **New Features**
  - Added AWS permissions linking through Amazon Cognito from Profile Integrations.
  - Added connection status, email, timestamps, and disconnect controls.
  - Added Cognito configuration support through settings.
  - Added secure token handling and validation for identity connections.

- **Bug Fixes**
  - Updated Cognito connection and callback documentation and test routes to use the correct API paths.

- **Tests**
  - Added coverage for connection flows, token validation, storage, revocation, configuration, and interface copy.

### DIFF
--- a/docs/superpowers/plans/2026-08-28-per-user-permissions-5b-connect-aws-identity.md
+++ b/docs/superpowers/plans/2026-08-28-per-user-permissions-5b-connect-aws-identity.md
@@ -699,7 +699,7 @@ Two endpoints mirroring the `/azure/connect` and `/azure/callback` pair that alr
 
 **Interfaces:**
 - Consumes: `CognitoSettings` (Task 1), `AwsIdentityLinkStore` (Task 3).
-- Produces: `GET /api/cloud-identity/cognito/connect` → 302 to the pool's authorize endpoint; `GET /api/cloud-identity/cognito/callback` → exchanges the code and stores the token.
+- Produces: `GET /api/v1/auth/cloud/cognito/connect` → 302 to the pool's authorize endpoint; `GET /api/v1/auth/cloud/cognito/callback` → exchanges the code and stores the token.
 
 - [ ] **Step 1: Read the Azure pair**
 
@@ -739,7 +739,7 @@ public class CognitoConnectEndpointTests(SharedWebAppFixture fixture)
             AllowAutoRedirect = false,
         });
 
-        var response = await client.GetAsync("/api/cloud-identity/cognito/connect");
+        var response = await client.GetAsync("/api/v1/auth/cloud/cognito/connect");
 
         response.StatusCode.Should().NotBe(HttpStatusCode.Redirect,
             "there is nowhere valid to redirect to");
@@ -755,7 +755,7 @@ public class CognitoConnectEndpointTests(SharedWebAppFixture fixture)
             AllowAutoRedirect = false,
         });
 
-        var response = await client.GetAsync("/api/cloud-identity/cognito/callback?code=abc");
+        var response = await client.GetAsync("/api/v1/auth/cloud/cognito/callback?code=abc");
 
         response.StatusCode.Should().NotBe(HttpStatusCode.OK);
     }
@@ -769,7 +769,7 @@ public class CognitoConnectEndpointTests(SharedWebAppFixture fixture)
         });
 
         var response = await client.GetAsync(
-            "/api/cloud-identity/cognito/callback?code=abc&state=not-a-state-we-issued");
+            "/api/v1/auth/cloud/cognito/callback?code=abc&state=not-a-state-we-issued");
 
         response.StatusCode.Should().NotBe(HttpStatusCode.OK);
     }
@@ -992,7 +992,7 @@ Expected: FAIL — the markup contains none of it.
 In the Cloud Identities section of `src/Connapse.Web/Components/Pages/ProfileIntegrations.razor`, add a card beside the existing AWS and Azure ones. It shows one of three states:
 
 - **Not configured** — Cognito settings are absent. Say an administrator sets this up, and do not offer a button that cannot work.
-- **Not connected** — a Connect button that navigates to `/api/cloud-identity/cognito/connect`.
+- **Not connected** — a Connect button that navigates to `/api/v1/auth/cloud/cognito/connect`.
 - **Connected** — the email it is connected as, when, and a Disconnect button.
 
 **Disconnect must revoke, not just forget.** Deleting the local row leaves the refresh token valid at Cognito, so anything that already copied it keeps working — the link is gone from Connapse's point of view and alive from AWS's. Call the pool's `POST {Domain}/oauth2/revoke` with the token and the client credentials **before** deleting the row, and delete the row whether or not revocation succeeded: a user who clicks Disconnect must end up disconnected locally regardless of what AWS says.
--- a/src/Connapse.Core/Models/AuthModels.cs
+++ b/src/Connapse.Core/Models/AuthModels.cs
@@ -119,5 +119,13 @@ public record AwsIdentityLinkDto(string Email, DateTime ConnectedAt, DateTime? L
 /// <see cref="Deleted"/> and <see cref="RevokedSuccessfully"/> are independent: the local row is
 /// removed regardless of whether Cognito could be told, so a caller can report "disconnected here,
 /// but AWS could not be told" instead of a uniformly clean success.
+/// <para>
+/// <see cref="LinkChangedDuringDisconnect"/> is a third, distinct outcome from a plain "nothing to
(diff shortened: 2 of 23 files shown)


Destrayon:
Retargeted to `epic/aws-per-user-permissions` rather than `main`, per [#436](https://github.com/Destrayon/Connapse/issues/436). Nothing in this epic reaches `main` until the AWS story is coherent end to end — specifically not until the original AWS device-flow sign-in has been removed (#435) rather than sitting on the integrations page beside its replacement.

Still mergeable, 8 commits, unchanged content.

Destrayon:
## Fixes Applied Successfully

Fixed 6 file(s) based on 5 CodeRabbit feedback item(s) — 4 applied, 1 declined.

**Commit:** `add8247` · Unit 1397 passed, integration 409 passed.

**Applied:**

- **Require HTTPS for the Cognito endpoint URLs.** The right call, and it corrects a reasoning error on my part: I put the loopback exception on the wrong thing. `IssuerUrl` and `Domain` are always AWS-hosted — there is no such thing as a localhost Cognito pool — so allowing `http://localhost` there bought nothing while widening what is accepted, and `TryRevokeAsync` POSTs the refresh token and client secret to `Domain`. The loopback allowance belongs to the OAuth **redirect URI**, which is Connapse's own address and is computed from the incoming request rather than read from these settings. Verified a localhost deployment is unaffected.

- **Disconnect can no longer delete a link it did not revoke.** `DisconnectAsync` fetched a link, revoked that token, then deleted by user id — so a reconnect landing in between meant the **new** link was deleted while its refresh token stayed live at Cognito, which is precisely what disconnect exists to prevent. An `Id` check would not have caught it, since `SaveAsync` updates in place. Delete is now conditioned on the protected token still matching what was revoked, with a distinct `LinkChangedDuringDisconnect` outcome when it does not.

- **Concurrent connects no longer race the unique index.** `SaveAsync` catches the Postgres unique violation on insert and retries as an update. A raw-SQL `ON CONFLICT` upsert would read more directly, but the store's unit tests run against EF Core's InMemory provider, which cannot execute raw SQL — noted in case that trade looks wrong.

- **A cancelled consent no longer returns a bare 400.** `code` and `state` were non-nullable query parameters, so Cognito's `error=access_denied` redirect (which carries no `code`) was rejected by model binding before the handler ran: a raw error page instead of the `/profile/integrations` redirect every other failure uses, with the three flow cookies left in the browser. Now nullable, with the provider error mapped to one of our own fixed reasons rather than echoed.

**Declined:**

- **Make `CognitoSettings` a record.** `CLAUDE.md` does say "records for DTOs and settings models", but `AwsSsoSettings` and `AzureAdSettings` are both classes. Making one of three siblings a record is worse than three consistent classes, and changing all three would touch an Azure path, which this epic explicitly forbids. Worth revisiting when Azure support lands and the convention can be applied uniformly.

**Still open, and disclosed rather than fixed:** the insert-vs-insert race is closed, but two simultaneous reconnects that both find an existing row have no ordering guarantee beyond commit order. Not exception-prone, and the loser's token is simply superseded.

Issue #435, opened 2026-08-28
Remove the AWS device-flow sign-in: it feeds nothing and promises something it cannot deliver


Destrayon:
The AWS IAM Identity Center device-flow link on the integrations page, and its Sign-in requirement on the providers page, should both go.

## Why

**It feeds nothing.** The only consumer of a stored cloud identity would be `CloudScopeService`, whose sole reference in the codebase is its own DI registration in `Program.cs`. Zero production callers — the finding that opened [#421](https://github.com/Destrayon/Connapse/issues/421).

**Its promise is now known to be unkeepable.** The card says Connapse stores this "for future per-user search filtering". A spike against a live AWS account proved that future cannot arrive by this route: `sso-oidc:CreateToken` has no `awsAdditionalDetails` field at all, and `PutApplicationGrant` excludes `device_code`, so a device-flow token can never carry an identity context. Per-user permissions arrive through the Cognito link added in #433 instead.

**Nobody has used it.** `user_cloud_identities` is empty.

So it stores data nothing reads, to enable a feature it structurally cannot enable, next to a card that actually delivers it.

## Scope

Remove:

- The **Amazon Web Services (Sign-In)** card on `/profile/integrations`
- `/aws/device-auth` and `/aws/device-auth/poll` endpoints, and `StartAwsDeviceAuthAsync` / `PollAwsDeviceAuthAsync` / `IsAwsSsoConfigured` on `ICloudIdentityService`
- `IAwsSsoClientRegistrar` and `AwsSsoClientRegistrar`
- `AwsSsoSettings`, `AwsSsoSettingsTab.razor`, `AwsSsoConnectionTester`, and the `awssso` settings category
- The AWS **Sign-in** requirement in `ProviderSetupReader` and its rendering on the providers page — an admin must not be asked to configure something that no longer exists
- `docs/aws-sso-setup.md`
- The AWS-specific assertions in `CloudIdentityClaimsTests`, `CloudIdentityServiceTests`, `ProvidersPageTests`, `ProviderSetupReaderTests`, `NewSettingsCategoriesIntegrationTests`

Keep, and do not confuse with the above:

- **Everything Azure.** The card, `/azure/connect`, `/azure/callback`, `AzureAdSettings`, its providers-page requirement. Azure support is coming.
- `UserCloudIdentityEntity` and the `user_cloud_identities` table — Azure stores there.
[...]

Issue #436, opened 2026-08-28
Epic: per-user AWS search permissions


Destrayon:
Integration branch for the AWS half of [#421](https://github.com/Destrayon/Connapse/issues/421). **Every PR in this epic targets `epic/aws-per-user-permissions`, not `main`.** Nothing reaches `main` until the AWS story is coherent end to end — in particular, not until the original AWS sign-in has been replaced rather than sitting alongside its successor.

Azure is explicitly out of scope. No PR here touches an Azure path.

## The goal, unchanged

An administrator decides in AWS which documents each user may read. Connapse honours that decision when the user searches. It never authors a permission, never stores a copy of one, and never decides anything.

## What is already on `main`

| | |
|---|---|
| Phases 1-4 | The enforcement seam: `ISearchScopeResolver`, `SearchScopes`, a prefix predicate in both search stores |
| 5a (#427) | Enforcement made correct: grant-scope normalisation, object grants matched by equality, the ways a search reaches nothing kept apart, failing closed, and the null-coordinate rule inverted after testing showed the old one would have hidden 127,892 of 127,898 documents on a real deployment |
| Design + research | `docs/superpowers/specs/2026-08-27-per-user-aws-search-permissions-design.md`, and a research report recording two AWS paths tested against a live account and found closed |

## What lands on this branch

- [x] **#433** — connect an AWS identity through Cognito. Already open as #434, retargeted here.
- [ ] **#435** — remove the AWS device-flow sign-in, from the integrations page and the providers page. It feeds nothing, and a spike proved it can never carry an identity context.
- [ ] **5b-ii** — keep the connection alive: refresh, the weekly touch job that stops idle expiry, and surfacing a link that has broken. "Permanent" is a property the feature acquires here.
- [ ] **5c** — provider setup and detection: the CloudFormation stack and CLI script an admin runs, and a page that reports honestly whether it worked. Must check application assignment explicitly — a missing one is the only failure in the chain that reports nothing useful.
- [ ] **5d** — the resolver. `ListCallerAccessGrants` behind `ISearchScopeResolver`, at which point filtering actually happens.

## Constraints carried across the epic

**The 5d resolver must never return `Unrestricted`.** `ScopeResolution.Guard` passes that value through for a caller with no user, because today only the default resolver produces it. A real resolver returning it on any path hands everything to an unidentified caller, silently.

[...]

PR #437, opened 2026-08-28
refactor: remove the AWS device-flow sign-in

Destrayon:
## What

Removes the AWS IAM Identity Center **device-flow sign-in** — the card on `/profile/integrations` and its Sign-in requirement on the providers page. 31 files, **+45 −2581**.

Targets `epic/aws-per-user-permissions`, not `main`, per [#436](https://github.com/Destrayon/Connapse/issues/436).

## Why

Closes #435

Three reasons, and the second is the one that settles it:

**It fed nothing.** The only consumer of a stored cloud identity would be `CloudScopeService`, whose sole reference in the codebase was its own DI registration — the finding that opened [#421](https://github.com/Destrayon/Connapse/issues/421).

**Its promise could not be kept.** The card said Connapse stored this "for future per-user search filtering". A spike against a live AWS account proved that future cannot arrive by that route: `sso-oidc:CreateToken` has no `awsAdditionalDetails` field at all, and `PutApplicationGrant` excludes `device_code`, so a device-flow token can never carry an identity context. Per-user permissions arrive through the Cognito link merged in #434 instead, which leaves this card promising, next to a card that delivers, something it structurally cannot do.

**Nobody had used it.** `user_cloud_identities` was empty.

## What went

The card and its handlers; `/aws/device-auth` and `/aws/device-auth/poll`; `StartAwsDeviceAuthAsync` / `PollAwsDeviceAuthAsync` / `IsAwsSsoConfigured`; `IAwsSsoClientRegistrar` and `AwsSsoClientRegistrar`; `AwsSsoSettings`, its settings tab, its connection tester, and the `awssso` category; the AWS Sign-in requirement in `ProviderSetupReader` and on the providers page; `AwsIdentityProvider`; `AwsDeviceAuthStartResult`; `docs/aws-sso-setup.md` and every link to it.

## What deliberately stayed

**All of Azure** — card, endpoints, settings, providers-page requirement, `AzureIdentityProvider`. Azure support is coming, and no Azure path is touched by this PR. The only Azure-adjacent edit in the diff is a `<see cref>` in `CognitoSettings.cs` repointed from the deleted `AwsSsoSettings` to `AzureAdSettings`.

**The AWS *connector* identity** — `ConnapseAwsCredentials`, `IS3Discovery`, `S3Discovery`, `AwsIamUserSetup`, `S3SetupPolicy`. These read "AWS" but are Connapse's own reader credential for S3 sources, unrelated to per-user sign-in. This naming overlap was the likeliest way the removal could go wrong; it didn't.

**`UserCloudIdentityEntity` and its table**, because Azure stores there — so **no migration**. Also `ICloudIdentityProvider`, `CloudScopeService` and the `CloudProvider` enum: dead, but symmetric across both clouds, and deleting half leaves something stranger than deleting none. `AwsIdentityProvider` is the exception, and the distinction is deliberate — it is not half of a pair, it is the half whose input can no longer be created.

## Type of Change

- [x] Refactoring (no functional changes)
- [x] Documentation update

## Testing

- [x] Unit tests pass — **1349 passed** (from 1397; −48)
- [x] Integration tests pass — **406 passed** (from 409; −3)

Every removed test is accounted for and was deleted because its subject went away, not because it broke: `AwsSsoSetupTests` (35), `AwsIdentityProviderTests` (3), the `IsAwsSsoConfigured` and device-auth regions of `CloudIdentityServiceTests` (7), three `InlineData` cases in `ProvidersPageTests`, and integration-side `NewSettingsCategoriesIntegrationTests` (2) plus one `CloudIdentityEndpointTests` case. Build is 0 errors **and 0 warnings** — a warning is how a half-finished removal announces itself.

## Both pages still read coherently

**Providers** — the Sign-in heading and its paragraph moved inside the Azure branch rather than rendering unconditionally with an inner AWS/Azure split, so `/providers/aws` has no empty section or dangling heading. AWS is still on the list page for its connector setup; its blurb went from "Sign in with AWS, and read files from S3." to "Read files from S3."

**Integrations** — the intro said "The AWS and Azure sign-in cards below", plural, describing two cards where one now remains. Rewritten to describe what is actually there: one Azure sign-in card storing metadata only, and one AWS Permissions card storing an encrypted refresh token. The contrast between those was the sentence doing real work, and it survives, as does the warning that linking does not yet filter search results.

## One thing worth knowing

The stalest reference was not in `src/` — `.claude/skills/release-gatekeeper/references/api-surface.md` still listed the removed endpoints and settings category, and that file is the checklist the release-validation skill drives. It would have produced a failing release run rather than a confused reader. Also cleaned: three broken links to the deleted doc in `README.md`, `docs/deployment.md` and `release/README.md`, and an `.env.example` block advertising variables that bound to nothing.


Destrayon:
Addressed the Codex adversarial review in 26845b4 — docs only, no code touched.

All three findings confirmed and fixed:

- **SECURITY.md** — the AWS device-flow setup path is gone, and the "no stored AWS access keys" promise was false: `AwsIamUserSetup` generates an IAM access key and `PostgresProviderCredentialStore` persists it. Replaced with what actually happens, including the DataProtection purpose.
- **Release gate** (`.claude/skills/release-gatekeeper/references/test-checklist.md`) — section 11.1 replaced with Cognito checks. Note it says to configure over `PUT /api/settings/cognito`, because Cognito has no admin UI card yet; a check for one would have failed.
- **5b-i plan** — marked complete (delivered by #434) and the dead `"awssso"` reference repointed.

**One thing the review missed, worse than what it found.** SECURITY.md and README.md both still claimed IAM-derived scope enforcement filters search — "cloud permissions are the source of truth", including in the repo's headline feature bullet. `CloudScopeService` has no production caller. #422 corrected this in `architecture.md` but missed both of these files, so the security policy has been promising an enforcement boundary that does not exist. Now stated plainly in a new "Corrections since v0.3.0" section.

**Deliberately not fixed here:** `docs/azure-identity-setup.md` carries the identical false claim, including "Search (scope filter injected as path prefix)". Left alone because of the standing instruction not to touch Azure paths in this epic — tracked separately.

Scope note: the AWS credential-storage correction is pre-existing and not caused by this branch, but it sits on a line this branch had to edit, so I fixed it rather than leave a knowingly false security claim behind.

CHANGELOG.md is untouched — its v0.3.1 entry is an accurate record of what shipped.

Destrayon:
Superseded by #447. Every commit on this branch is already an ancestor of `feature/445-identity-center-saml-signin`, so the device-flow removal ships as part of that PR rather than separately.

An adversarial review was run against this branch in isolation before closing. It flagged that the per-user AWS trust boundary is deleted here with no replacement — which is accurate for this diff read against the epic base, and resolved further along the branch: the SAML sign-in and `AwsSearchScopeResolver` provide the replacement, and it now fails closed rather than falling back to the deployment identity.

The same review noted stranded AWS SSO settings after the code removal. That one is real and confirmed against a live database; tracked separately in #448.

Issue #438, opened 2026-08-28
Configure Cognito from the admin UI, and only from there


Destrayon:
Part of #436.

## The problem

`CognitoSettings` shipped in #434 with a settings category and read/write arms on `/api/settings/{category}`, but no admin form. The only way to configure a Cognito pool today is `PUT /api/settings/cognito`.

That is backwards for this deployment. Every other identity provider on the Providers page is configured through a form; Cognito should be too, and the API should not be a second way in.

It is also visibly broken as a user journey: `ProfileIntegrations.razor` tells the administrator to "set up the pool under Settings", and there is nothing under Settings to find.

## What to do

1. Add a Cognito settings form to the AWS provider page (`Providers.razor`, `Key == "aws"`), following the shape of `AzureAdSettingsTab`. Fields: Issuer URL, Domain, Client ID, Client Secret, Region.
2. Remove the `"cognito"` arms from `SettingsEndpoints.cs` — both the GET and the PUT — so the admin UI is the only way to configure it.
3. Point the integrations-page copy at the real location.
4. Update the release checklist, which currently says to configure Cognito over the API.

## Notes

- `Providers.razor` writes through `ISettingsStore` directly, not through the API, so removing the endpoint arms does not affect the form.
- The integration tests in `CognitoConnectEndpointTests` already save through `ISettingsStore`, so they are unaffected.
- Removing the GET arm also stops the pool's client secret being readable over the API, which is worth having on its own.
- Out of scope: a Cognito connection tester. Worth considering later — a wrong Issuer URL currently fails much later at token exchange with an error that names neither side — but it is not needed to close this.

PR #439, opened 2026-08-28
Set up Cognito and Identity Center from the admin UI

Destrayon:
Closes #438. Part of #436.

> **Stacked on #437.** Branches off `feature/435-remove-aws-device-flow`, so the diff also carries its files until #437 merges.

## Why

`CognitoSettings` shipped in #434 with a settings category and no admin form, so the only way to point Connapse at a pool was `PUT /api/settings/cognito` — and `ProfileIntegrations.razor` told administrators to set it up "under Settings", where there was nothing to find.

That was the starting point. Configuring a pool by hand turned out to need eleven values an administrator has no way to invent, so most of this branch is the guided setup that produces them, and the corrections that came out of running it end to end against a live account.

## What this adds

**A form, and only a form.** `CognitoSettingsTab.razor` on the AWS provider page. Both `cognito` arms are gone from `SettingsEndpoints.cs` — the GET was returning the pool's client secret to anyone holding an admin token.

**Guided setup in three steps**, each with a read-first script and a paste-back:

1. **IAM Identity Center** — a read-only scan that finds the instance and, importantly, its region. Identity Center lives in exactly one region per organisation and nothing else an administrator holds encodes which; the Cognito script previously looked only in CloudShell's own region, where an instance one region over reads as no instance at all.
2. **Per-user permissions** — a CloudFormation template plus the four `sso-admin` calls that have no CloudFormation resource type.
3. **Federation** — the SAML application, which cannot be created in one pass because its audience is built from a pool id Cognito only generates once the pool exists.

**The join key is the Identity Center user name, not the email.** A trusted token issuer resolves one claim against one of three identity-store attributes, and email is the one that cannot work: Cognito marks a federated user's mapped address unverified and cannot verify it, so the validator refused every sign-in the design depends on. Moving the key removed the need to weaken that check rather than weakening it.

**Federation is required.** A pool that federates nowhere holds only accounts made by hand, and each still needs a matching Identity Center user to resolve against — two accounts per person, the second doing the actual work.

## Verified end to end

Against account `086015909943`: Identity Center authenticates, Cognito converts the SAML assertion to an OIDC token, Connapse validates it and resolves `diviel`. The link is stored with the directory user name rather than an address.

Six failures on the way, each now covered by a test or a check:

| Failure | Cause |
|---|---|
| `invalid_request` on authorize | asked for `offline_access`, which Cognito has no such scope for |
| CloudShell session ending | `set -e` and `exit` in a script pasted into a live shell |
| copy blocked as a ClickFix attack | the fix for that had wrapped it in `bash -s <<...` |
| `The provider Workforce does not exist` | the client named the provider as a string, so CloudFormation was free to build it first |
| settings block printed after a rollback | removing `set -e` also removed the only check on the deploy |
| `AWS did not say which directory user you are` | `preferred_username` is a `profile` claim and the scope asked for `openid email` |

## What this does not do

No access grants are created — who may read what stays the administrator's decision, authored in AWS. Nothing consumes the identity link yet: search is unfiltered for everyone until the resolver in 5d exists, which the profile card says plainly.

The link lasts 30 days. `RefreshTokenValidity` is 30 and rotation is not configured, and Cognito's clock runs from issuance rather than last use — so a refresh job alone would not extend it. That is worth settling before 5b-ii is built, because the item as written would not have worked.

## Tests

1009 Core, 108 Identity, 82 Web, 410 Integration.


Destrayon:
Superseded by #447. Every commit on this branch is already an ancestor of `feature/445-identity-center-saml-signin`.

Worth recording why this one is closed rather than merged: it is self-superseding. It adds the Cognito admin UI, and a later commit on the branch that replaced it (`87d006a`) removes Cognito entirely in favour of signing in through Identity Center over SAML directly. Merging it would add a feature the same epic then deletes.

The settings rows Cognito left behind are tracked in #448.

Issue #451, opened 2026-09-01
Concurrency-safe SAML/group resets on the AWS providers page


Destrayon:
## Problem

On the AWS providers page, `ClearGrantGroup` and `ClearSamlApplication` (`Providers.razor`) rebuild the entire `SamlSignInSettings` from the page's `samlSettings` field — a snapshot captured once in `OnInitialized` from `SamlOptions.CurrentValue`. On a long-lived page, if another administrator rotates the signing certificate / issuer (or changes the grant group) in the meantime, the reset writes the stale snapshot back, silently reverting that change (a lost update — potential auth outage / trust regression).

Flagged by a Codex adversarial review of PR #449.

## Fix

Build the reset's replacement settings from the freshest stored view at reset time — `SamlOptions.CurrentValue`, which reflects reloads from concurrent saves — instead of the page-load snapshot, and clear only the fields that reset owns. (Reading the DB row directly via `ISettingsStore.GetAsync` is deliberately avoided: an env-configured deployment has no row, and writing one would shadow the environment — the trap `SaveSamlSettings` already documents.)

Full optimistic-concurrency (reject-on-conflict) is out of scope: `ISettingsStore` has no version/etag primitive, so it would be a store-wide schema change. Re-reading the live value eliminates the load-time snapshot window, which is the actual defect.

## Acceptance

- Resetting the group preserves a concurrently-rotated certificate/issuer.
- Resetting the SAML application preserves a concurrently-changed grant group.
- Tests cover the reset transforms and that the resets read the live settings, not the snapshot.

PR #452, opened 2026-09-01
fix: concurrency-safe SAML/group resets (#451)

Destrayon:
## What

Make the two per-user-permission resets (`ClearGrantGroup`, `ClearSamlApplication`) build the replacement `SamlSignInSettings` from the **live** monitor value (`SamlOptions.CurrentValue`) instead of the `samlSettings` snapshot captured once in `OnInitialized`.

## Why

Closes #451. On a long-lived page, another administrator can rotate the signing certificate/issuer or change the grant group. The old resets rebuilt the whole settings object from the page-load snapshot and saved it, silently reverting that concurrent change (a lost update → auth outage / trust regression). Flagged by a Codex adversarial review of #449.

## How

- Extracted two pure helpers, `WithoutGrantGroup` and `GrantGroupOnly`, that clear only the fields their reset owns.
- Both resets now pass `SamlOptions.CurrentValue` to those helpers, so a reset operates on the freshest stored view (which reflects reloads from concurrent saves) and preserves whatever the other fields currently hold.
- Reading the DB row directly (`ISettingsStore.GetAsync`) is deliberately avoided — an env-configured deployment has no row, and writing one would shadow the environment (the trap `SaveSamlSettings` documents). Full optimistic concurrency (reject-on-conflict) would need a store-wide version primitive and is out of scope; re-reading the live value closes the load-time snapshot window, which is the actual defect.

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [x] Test update or addition

## Testing

- [x] Unit tests pass (`dotnet test --filter "Category=Unit"`)
- [x] Integration tests pass (`dotnet test --filter "Category=Integration"`)

New `SamlResetConcurrencyTests` (5): both transforms preserve the fields they must and drop the ones they must; a rotated-vs-stale certificate scenario; and a source-level guard that both resets read `SamlOptions.CurrentValue`. Full `Connapse.Web.Tests` (111) and the integration rendering tests pass.

## Notes

A faithful end-to-end two-session render test would require a Blazor component renderer (bUnit / the full dependency graph), since the reset chains into `StateHasChanged`. Coverage here is the pure transforms plus a wiring guard, which is proportionate to the fix.

## Checklist

- [x] Under 300 lines
- [x] Does one thing
- [x] Self-reviewed
- [x] Tests added
- [x] No new warnings or errors

### DIFF
--- a/src/Connapse.Web/Components/Pages/Providers.razor
+++ b/src/Connapse.Web/Components/Pages/Providers.razor
@@ -1317,11 +1317,11 @@
         // incomplete, which the scope resolver reads as EnforcingButUnusable and denies — searches
         // return nothing for everyone until the application is set up again, rather than silently
         // widening the whole corpus to every user.
-        await SaveSamlSettings(new SamlSignInSettings
-        {
-            GrantGroupId = samlSettings.GrantGroupId,
-            GrantGroupName = samlSettings.GrantGroupName,
-        });
+        //
+        // Built from CurrentValue, not the samlSettings snapshot captured at page load: on a
+        // long-lived page another administrator may have changed the group since, and rebuilding
+        // from the stale copy would write that obsolete value back.
+        await SaveSamlSettings(GrantGroupOnly(SamlOptions.CurrentValue));
 
         if (saveSuccess)
         {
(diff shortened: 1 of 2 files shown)


Destrayon:
A Codex adversarial review raised two points on this branch, both considered and consciously deferred:

1. **Residual TOCTOU** — reading `CurrentValue` narrows the lost-update window (minutes → milliseconds) but doesn't close it; `SaveAsync` is still a blind full-row overwrite. Fully closing it needs store-wide optimistic concurrency (row version + conditional save + retry), which touches every settings category. Deferred as disproportionate to the risk: the race is now sub-second, admin-only, and fails **loud + fail-closed** (a reverted cert makes searches deny, not widen). Tracked in #453.
2. **Tests prove field-copying + wiring, not concurrency** — a two-writer interleaving test only has something to assert once real conflict-detection exists. Also tracked in #453.

This PR stands as the incremental improvement (removes the realistic minutes-long window, no regression).

Issue #503, opened 2026-09-09
Audit Azure provider, Blob connection, and source config pages against the config-page rubric


Destrayon:
Same rubric as #468, applied to the Azure epic (`epic/azure-blob-provider`): the Azure provider page (Access and Per-user permissions cards, easy and manual paths), the Azure Blob connection form, the Azure source branch, the Entra card on Integrations, `docs/azure-setup.md`, and the code behind them.

Baseline: all five gates pass. Findings to fix:

- **New certificate overwrites the live key before Save.** `LoadOrMintAzureCertificate(newCertificate: true)` writes straight to the path the stored settings point at, so an already-working Access step breaks on the next restart without confirmation. Mint to a pending file; promote on Save.
- Access status is config-only; Recheck re-reads settings rather than asking Azure. Add a live token probe so an expired or unregistered certificate reads as Failed with the AADSTS reason, not Ready.
- Manual fields have no where-to-obtain links; sign-in redirect placeholder shows `/azure/cb` (the real callback is `/azure/callback`); tenant and redirect URI are not pre-filled on the manual sign-in form.
- Config-key names (`Providers:Azure`, `Identity:AzureAd`, "RBAC resolver") in user-facing requirement text.
- Certificate expiry (1-year self-signed) never shown; no warning before it lapses.
- Test-connection messages do not say which layer failed; auth failures fall through to the raw exception.
- No "Add a connection" action once Access is Ready; page has no link to the setup doc; doc lists no manual fields.
- Mixed required/optional conventions; "Re-check" vs "Recheck"; Integrations card points at "Providers → Microsoft Entra ID" (the page is "Azure").
- Stray developer text in `ProviderSetupReader`, `CloudIdentityEndpoints`, `Providers.razor`.

Report goes to `docs/plans/azure-config-page-audit-report.md`. Loop ends when an adversarial review passes the epic changes.

PR #504, opened 2026-09-09
fix: Azure config pages audited against the config-page rubric

Destrayon:
Closes #503.

Audits the Azure provider page (Access and Per-user permissions cards, easy Cloud Shell path and manual forms), the Azure Blob connection form, the Azure source branch, the Entra card on Integrations, and `docs/azure-setup.md` against the config-page rubric, and fixes what it found. All five rubric gates passed from the start; the dimension score moved from 1.93 to 2.51 of 3. Eight Codex adversarial review rounds and one Claude review followed; every finding was verified against the code and fixed. Full report with per-criterion scores and evidence: `docs/plans/azure-config-page-audit-report.md` (local, gitignored) — summary below.

## Certificate lifecycle

- **"New certificate" no longer overwrites the live private key.** Each save attempt writes its candidate to its own thumbprint-and-attempt-named file under `appdata/azure`, verifies it with Entra from there, and only then commits settings naming it. The key in use is never moved or overwritten.
- Superseded files are swept only when nothing stored names them (both Azure settings categories re-read from the store) and they are over an hour old, so two processes sharing a volume cannot delete each other's key. If the store cannot be read, nothing is deleted.
- Both Cloud Shell scripts print the registered certificate's thumbprint; Save refuses a paste whose thumbprint is not the material the page holds (a second tab made a new one). Writes are atomic (temp + move); interrupted-write temp files are swept.
- Both cards show certificate expiry with a 30-day warning and an expired state; a missing or unreadable certificate file is an alert on the card, never an exception from the page.

## Live health

- `IAzureBlobDiscovery.CheckAccessAsync` acquires an ARM-scope token with a fresh credential each time (no cached token can mask a revoked certificate). The reader reports Ready with what was verified, Failed with the AADSTS reason when Entra refuses the credential (allowlisted codes only — throttling and transient faults read as Unconfirmed), Failed when the identity is unusable on this host, and Unconfirmed when Azure cannot be reached. The sign-in application's credential is checked the same way, in parallel with Access.
- Every manual or script save is checked with Entra before it is stored; a refused, unusable, or unconfirmable candidate leaves the working settings untouched and keeps what was typed.
- Both Azure cards render the reader's detail text with alert/status semantics, not just a status word. The connection tester names the layer that failed (reached / Entra accepted / allowed / exists), distinguishes refusal from transport failure, validates the blob endpoint, and puts the raw error in `Details`.

## Fail-closed enforcement (pre-existing epic bugs)

- The AWS sign-in save rewrote the enforcement record with only its own flag, switching Azure filtering off. Enforcement latches now merge with a monotonic OR through a new row-locked `ISettingsStore.UpdateAsync` (`SELECT … FOR UPDATE` in one transaction), are written **before** the sign-in settings, and must be visible in the live options before sign-in saves or the startup migration completes. The reader reports Per-user permissions as Failed if sign-in is configured while the latch is off.
- A user-assigned managed identity with the tenant recorded was rejected by `AzureCredentialChainFactory`; it is now an explicit mode the form, factory, and discovery agree on.
- `SaveSettings` returns committed / not committed / unknown; an unknown outcome never rolls back a certificate. `PostgresSettingsStore` clears its change tracker on a failed write so a failed rotation cannot ride along with the next save in the same request.

## Forms and copy

- Access manual form: certificate-app vs user-assigned managed identity as a `fieldset` radio revealing one branch (the other cleared on save); prerequisites stated; a portal link on every field; gathering order (tenant → identity → subscription, optional); submit validation with `role="alert"` and `aria-invalid`.
- Sign-in manual form: tenant and redirect URI pre-filled, correct callback placeholder (`/api/v1/auth/cloud/azure/callback`), links, validation.
- One "(optional)" convention; config keys (`Providers:Azure`, `Identity:AzureAd`, "RBAC resolver") out of user copy; Graph permissions named at the consent step; "Recheck" and "Add a connection" parity with AWS; Integrations card names the page that exists; connection Name/probe help cover containers; stale developer notes removed; a checking indicator instead of unmounting the cards on refresh.
- `docs/azure-setup.md`: field tables for both cards and the connection, what Recheck verifies, certificate lifetime and rotation, and troubleshooting entries matching every new message.

## Tests

Unit: Web 134, Storage 264 (incl. new `AzureCertificateFileTests`, discovery candidate/classifier tests, factory managed-identity tests), Core 1,117 (Cloud Shell thumbprint parse tests; the CRLF-sensitive subshell assertion now normalises line endings), Identity 109, Ingestion 219, Background 18 — all green. Integration (Testcontainers PostgreSQL): new `SettingsStoreUpdateIntegrationTests` (twelve concurrent scopes merging end with both flags on), plus the settings, provider-page, and cloud-identity suites — green.

Manual checks still recommended: tab order and focus under the sticky header on `/admin/providers/azure`, `.form-text` contrast on both themes, 200% zoom, and the live Entra check against a real tenant.

### DIFF
--- a/docs/azure-setup.md
+++ b/docs/azure-setup.md
@@ -61,11 +61,17 @@ not found (`AADSTS700016`, `AADSTS90002`), missing service principal (`AADSTS700
 a refusal. Throttling, transient faults, and outages read as *Unconfirmed* with a retry, never as
 a demand to set up again.
 
-Old certificate files are removed only once nothing stored names them and they are more than an
-hour old, so two Connapse processes sharing one volume cannot delete each other's key. Saving
-the sign-in application switches per-user enforcement on *before* storing the application; if
-the switch cannot be written the application is not saved, and if it is somehow off while sign-in
-is configured the Per-user permissions card shows Failed and says how to switch it on.
+Each save attempt writes its own file (`connapse-azure-access-<thumbprint>-<attempt>.pem`), so
+attempts never share one. Old certificate files are removed only once nothing stored names them
+and they are more than an hour old, so two Connapse processes sharing one volume cannot delete
+each other's key; if the settings store cannot be read, nothing is deleted at all. Saving the
+sign-in application switches per-user enforcement on *before* storing the application and
+requires the switch to be live in this process; if it cannot be written or applied, the
+application is not saved and the message says to restart and save again. The AWS sign-in save
+carries the Azure switch through unchanged. If enforcement is somehow off while sign-in is
+configured, the Per-user permissions card shows Failed and says how to switch it on. A save
+whose outcome cannot be confirmed (the store failed after committing and could not be read back)
+is reported as unconfirmed rather than as failed; check the card after a restart before retrying.
 
 ## Manual values
 
--- a/docs/azure-setup.md
+++ b/docs/azure-setup.md
@@ -32,6 +32,21 @@ from the lists (or type them), Test connection, Save. People link their Entra id
 Integrations → Microsoft Entra ID; their search results are then limited to what that identity
 may read.
 
+### Connapse running on Azure
+
+If the Connapse host is an Azure VM, App Service, or container with a managed identity, the
+Access guide notices and switches to it: no app registration and no certificate. Its script
+creates the same two custom roles and assigns them to that identity by object id, and the
+paste-back records the identity (tenant, subscription, principal). Whoever runs it needs only
+Owner or User Access Administrator on the subscription. A link under the guide switches back to
+a certificate app if you prefer one; the manual form also offers "This host's managed identity"
+as a choice, needing only the tenant.
+
+The Per-user permissions guide then grants the identity its two Graph permissions as app roles
+(a REST call per permission) instead of consenting an app. If the person running it is not a
+Global Administrator or Privileged Role Administrator, the script reports that and the card
+shows the two commands to hand to one; press "The permissions have been granted" once they have.
+
 ### What Recheck verifies
 
 **Recheck** on the Access card asks Entra for a token as Connapse's identity. *Ready* means Entra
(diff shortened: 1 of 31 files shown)


Destrayon:
Added in `44f453f`: **guided setup uses the host's managed identity when Connapse runs on Azure.**

- On first render the Azure provider page asks the instance metadata service for a token; if the host has a managed identity, the token's `tid`/`oid`/`appid`/`xms_mirid` claims name it (and its subscription), and the Access guide switches to a script that creates the two custom roles and assigns them to that identity by object id — no app registration, no certificate, nothing to renew. Off Azure the probe fails within about a second and nothing changes. A link switches to a certificate app instead.
- The paste-back stores an explicit `UseHostManagedIdentity` flag plus the principal, tenant, and subscription, after a fresh-credential check; the flag is what keeps "the host's identity" distinct from "nothing configured" for the fail-closed rules. `AzureCredentialChainFactory`, `AzureBlobDiscovery.IsConfigured`, and the reader all recognise it.
- The Per-user permissions guide grants the identity `User.Read.All` and `GroupMember.Read.All` as app roles by REST (there is no app to consent to); if the runner is not a Global/Privileged Role Administrator the card shows the two commands to hand to one.
- The manual Access form offers "This host's managed identity" as a third choice needing only the tenant.

New service `IAzureHostIdentity` / `AzureHostIdentity` (singleton, caches a found identity). Tests cover the chain, the configured rule, the claim parser and cache, both scripts and parsers, and the permissions branch; the provider-page render and Azure DI integration tests pass.

**Not yet exercised against a real Azure host.** The detection, the object-id role assignment, and the Graph app-role REST grant follow Microsoft's documented behaviour; one live run on an Azure VM or App Service with a managed identity is the remaining verification.

Destrayon:
Follow-up in the next commit: a Claude adversarial review of the managed-identity commit returned FAIL with four blockers, all fixed — a recorded principal could be carried onto a different user-assigned identity (sending the permissions script's Graph grants to the wrong principal); the "use the host's identity instead" link did nothing when a certificate app was stored; the permissions script was not regenerated after Access changed kind; and a host identity entered by hand dead-ended the permissions guide. Also taken: detection only on the Azure page and concurrent with the status read, a stored managed identity on a host without one reads as Failed/unusable, naming both identities at once fails closed, GUID-only grant commands, and two copy fixes. All unit suites green.

Destrayon:
Claude adversarial review of the managed-identity work: **PASS** on round 6 (commits `44f453f` → `412c4ad`). Rounds 2–5 each found one blocker, all fixed on the branch:

- a hand-entered identity was pointed at a doc section that did not exist (now `docs/azure-setup.md` → "Graph permissions for a managed identity", linked from the page);
- detection ran only on first render, so arriving from the provider list never detected (now from `OnParametersSet`, non-blocking, with a checking line);
- a permissions script generated during detection kept the recorded principal (now rebuilt when detection lands);
- after a Recheck found a re-created identity, the Access script still named the old principal (now dropped and rebuilt).

Also taken along the way: tenant compared against the token's own `tid`; Recheck re-probes the host and forgets a removed identity; the managed-identity script's role update patches the permissions array so re-runs are not no-ops; GUID tenant required in the managed-identity modes; pending-permissions state always has a way out; several message and doc corrections.

Left as notes (non-blocking): a status line when a Recheck drops the Access script; keeping the last known identity on a transient metadata fault; normalising GUID tenant forms; a token seam for unit-testing the tenant check. Still not exercised on a real Azure host.

Issue #500, opened 2026-09-08
Azure Blob connector: guided setup at parity with S3 (account/container pickers, readiness, RBAC snippet)


Destrayon:
The Azure Blob connection form has two text fields and Test; the S3 form has a readiness banner, a bucket picker that fills the allowed list and name, region auto-detect, a generated least-privilege policy, and a troubleshooting link.

Research: docs/research/azure-blob-connector-parity-2026-09-08.md.

Plan:
1. Access script: add two read-only listing actions to the custom roles (`Microsoft.Storage/storageAccounts/read` on the subscription-scoped reader role, `.../blobServices/containers/read` on the blob data role), create-or-update so re-runs upgrade existing roles.
2. `IAzureBlobDiscovery` (Core) + `AzureBlobDiscovery` (Storage, Azure.ResourceManager.Storage): list storage accounts (blob endpoint, resource group, IsHnsEnabled), list containers, probe access — with an `AzureProbe<T>` outcome type mirroring `AwsProbe<T>`.
3. Connections form Azure block mirroring S3: readiness banner, account picker (Gen2 badge, manual fallback), container picker with name auto-fill, generated `az role assignment` snippet scoped to the chosen account/container, troubleshooting link.
4. docs/azure-setup.md troubleshooting section.

Read-only throughout; no keys, SAS, or connection strings. Part of epic #475.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #499, opened 2026-09-07
Surface Azure on the Providers admin page


Destrayon:
The admin Providers page (`/admin/providers`) lists only AWS: `ProviderSetupReader.ReadAsync` returns a hard-coded single `ProviderSetup("aws", …)`. The Azure backend (connections, per-user Entra link, permission engine — epic #475) is complete, but Azure never appears on this page, so operators can't see or set it up the way they can AWS.

Found while testing the epic before its PR to `main`. Fixed incrementally on `feature/azure-testing-fixes`.

## Increment 1 (this change)
Add an Azure `ProviderSetup` to `ProviderSetupReader` so Azure renders (via the page's generic provider list + detail requirements view), with honest config-derived status:
- **Access** — the Azure app credential (`Providers:Azure`: tenant + certificate app registration or managed identity).
- **Per-user permissions** — the Entra sign-in app (`Identity:AzureAd`) + the subscription the RBAC resolver queries (`Providers:Azure` SubscriptionId).

`InUse` = the installation has an `AzureBlob` connection. Config-only status for now (no live probe; an ambient system-assigned managed identity can't be seen from settings — noted in the requirement detail).

## Later increments (follow-ups)
- A live Azure access probe (parity with AWS's `WhoAmI`/bucket check).
- Guided setup UI on the Azure detail page (cert, subscription/tenant, Entra app-registration guidance, connection test) mirroring AWS's cards — currently Azure config is settings/env only.

Issue #490, opened 2026-09-06
Phase 4e — Gen2 live per-hit verify seam + tag verify (completes Phase 4)


Destrayon:
Sub-issue of #479 (Phase 4 of epic #475). Spec: `docs/superpowers/specs/2026-09-06-azure-phase4-permission-engine-design.md` §E. Milestone v0.4.0. Stacks on `epic/azure-blob-provider`. **Last sub-issue — on merge, epic → main.**

## Scope
- `ISearchResultVerifier` (Core) + no-op default (AWS-only). `HybridSearchService` calls it post-retrieval, **over-fetching + backfilling until the page is full or candidates exhausted**, bounded parallelism (`SemaphoreSlim` / `Parallel.ForEachAsync` ~16–64), SDK backoff honoring 503/Retry-After.
- Retrieve Gen2 by relevance over a safe **over-approximation** (exact RBAC/flat, else container-level/unfiltered) — folder structure never narrows retrieval.
- Routing per `azblob://` hit: exact RBAC scope → pass (RBAC supersedes ACLs); else HNS → verify the file's **own** ACL (4d) + ancestor traverse — **folder access never trusted** (a locked-down file beneath a readable folder is dropped; a bespoke per-file grant beneath an unreadable folder is returned); flat under a tag condition → tag verify (`GetBlobTags` / `FindBlobsByTags`). AWS / non-cloud → untouched.

## Acceptance
- [ ] **Soundness**: locked-down file beneath a readable folder is dropped (no over-grant)
- [ ] **Completeness**: a bespoke per-file ("Case C") grant beneath an unreadable folder is retrieved + returned; permission edits reflected within cache TTL
- [ ] Tag-conditioned grants verified live; exact-RBAC and AWS/flat hits skip verification
- [ ] Green build + tests; AWS path unchanged

Issue #489, opened 2026-09-06
Phase 4d — Gen2 POSIX ACL engine + ancestor-traverse resolver


Destrayon:
Sub-issue of #479 (Phase 4 of epic #475). Spec: `docs/superpowers/specs/2026-09-06-azure-phase4-permission-engine-design.md` §D. Milestone v0.4.0. Stacks on `epic/azure-blob-provider`.

## Scope (pure library)
- POSIX first-match evaluator: owner bypasses mask; named user/group capped by mask; any one group granting suffices; owning-group; other. Read (`R`) and traverse (`X`) decisions for identity set P.
- Ancestor-traverse resolver: given a file path, resolve whether P holds `X` on **every** ancestor directory; resolve lazily and cache per directory (mode bits via `GetPaths`/properties; `GetAccessControl` only for named-entry tie-breaks). Optional bulk pre-warm (one recursive directories-only `GetPaths`) — perf only, not required for correctness.
- ADLS ACLs are copy-at-create, not live-inherited: **folder access never proves file access** (neither direction). This engine underpins 4e; it never excludes and never skip-verifies from folders.

## Acceptance
- [ ] POSIX first-match matrix (owner/named/group/other + mask); ancestor-traverse (lose X on chain → unreadable); named-entry tie-break fallback; cache
- [ ] Green build + tests

PR #505, opened 2026-09-09
Azure Blob Storage as a per-object-permission provider

### LINKED ISSUE #475: Epic: Rebuild Azure Blob Storage as a per-object-permission provider
## Description
Rebuild Azure Blob Storage as a full provider mirroring the AWS per-user-permission work (epic #436), adapted to Azure's authorization model. Removes the existing unverified Azure Blob + Azure AD code and delivers: Connapse's own Azure app identity, an Entra user-identity link, and a per-object **query-time live** permission engine.

Design spec: `docs/superpowers/specs/2026-09-05-azure-blob-provider-design.md`.

**Governing principle — Universal Per-Object Permission Invariant:** every source is evaluated per object, at parity with the cloud's own authorization decision (never broader), failing closed on any uncertainty. Connapse only ever *reads* permissions the admin authored in Azure; it never writes grants.

## Phases (each is its own spec -> plan -> implementation cycle)
- [ ] Phase 1 — Teardown of the existing unverified Azure Blob + Azure AD code
- [ ] Phase 2 — `ConnapseAzureCredentials` + rebuilt data-plane connector/tester
- [ ] Phase 3 — Entra user identity link (OIDC oid/tid, permanent + revocable)
- [ ] Phase 4 — Per-object query-time live permission engine

## Key design decisions (see spec)
- **Query-time live** permission evaluation (not ingestion-time capture) so Connapse never drifts from Azure on ACL/role edits.
- **Flat accounts:** RBAC-only, container-scoped — resolve the searcher's scope set once, push prefixes into SQL.
- **Gen2/HNS accounts:** directory-subtree prefixes for recall (Phase A) + per-hit POSIX verification for parity (Phase B); `Storage Blob Data Reader` suffices.
- ARM `assignedTo('{oid}')` is transitive over groups (one call); `effective = grants - denies`.
- Composite two-stage resolver so AWS + Azure enforcement coexist; AWS path untouched.

## Related
Mirrors epic #436 (AWS per-user permissions).

Destrayon:
Closes #475. Also closes the phase and follow-up issues delivered on this epic: #477, #478, #479, #486, #487, #488, #489, #490, #499, #500, #501, #503.

Rebuilds Azure Blob Storage as a full provider mirroring the AWS per-user-permission work (epic #436), adapted to Azure's authorization model. Design spec: `docs/superpowers/specs/2026-09-05-azure-blob-provider-design.md`. Governing principle: every source is evaluated per object, at parity with the cloud's own authorization decision and never broader, failing closed on any uncertainty. Connapse only reads permissions the administrator authored in Azure; it never writes a grant.

## What ships

**Connapse's own Azure identity** (`a22b93f`) — a deterministic credential chain (certificate-authenticated app registration, a user-assigned managed identity, or the host's own managed identity; never a client secret, never a fall-through to a broader identity), a read-only Blob connector with a connection tester, and Data Lake Gen2 awareness.

**Entra user identity link** (`0cad97a`) — people link their Entra identity through an authorization-code + PKCE flow; the link records object id and tenant only, is confirmed in the user's own session (no cross-user linking), and is revocable.

**Per-object query-time live permission engine** (`a4e4807` → `9d8bb53`) — searcher identity resolved through Microsoft Graph (transitive group membership); RBAC scope resolution through Azure Resource Manager with ABAC condition support, deny assignments subtracted; flat accounts resolved once and pushed into SQL as prefixes; Gen2 accounts get directory-subtree prefixes for recall plus per-hit POSIX ACL and blob-tag verification for parity; a composite resolver so AWS and Azure enforcement coexist, each latched independently.

**Guided setup and connector parity** (`9295ef5`) — two Cloud Shell scripts (Access, Per-user permissions) that mint certificates on the Connapse host and upload only the public half, create two least-privilege custom roles, and register the sign-in app; storage-account and container pickers on the connection form; readiness on the Providers page; an RBAC snippet scoped to exactly the chosen containers.

**Config-page audit and hardening** (`ee1fa0a`, PR #504) — the Azure provider page, Blob connection form, source branch, Entra card, and setup doc audited against the config-page rubric (all gates pass; dimension score 1.93 → 2.51 of 3), then eight Codex and seven Claude adversarial review rounds with every finding fixed. Highlights:
- Certificate lifecycle is transactional: each candidate goes to its own thumbprint-and-attempt-named file, is verified with Entra from there, and settings are committed to name it before anything old is swept; the key in use is never moved or overwritten; expiry shown with a 30-day warning.
- Live health: Recheck acquires a token with a fresh credential; a refused certificate reads as Failed with the AADSTS reason (allowlisted codes only), an unusable identity as Failed, an unreachable Entra as Unconfirmed; both cards render the reason and the fix.
- Fail-closed enforcement: latches merge under a row lock (`ISettingsStore.UpdateAsync`), are written before sign-in settings, and must be visible in the live options before a save or the startup migration completes; the AWS sign-in save no longer resets the Azure flag; a managed identity with the tenant recorded is an explicit, recognised mode.
- When Connapse runs on an Azure host with a managed identity, the guided setup detects it and grants that identity the roles and the two Graph permissions directly — no app registration, no certificate.
- Forms: certificate app / host identity / user-assigned identity as a conditional reveal; a portal link and validation on every field; one optional convention; layered test-connection messages with the raw error on demand; `docs/azure-setup.md` with field tables, Recheck semantics, certificate lifetime, managed-identity grants, and troubleshooting matching every message.

## Tests

Unit: Web 142 (134 tagged), Storage 279, Core 1,126, Identity 109, Ingestion 219, Background 18 — all green. Integration (Testcontainers PostgreSQL, Azurite): connector, DI wiring for the provider, credential chain, directory readers, enforcement, cloud-identity endpoints, provider-page render, settings-store atomic update — green. CI runs on this PR; the epic's own branch-to-branch PRs did not run it.

## Not yet exercised live

The managed-identity guided path (detection from the instance metadata service, role assignment by object id, the Graph app-role REST grant) follows Microsoft's documented behaviour and has not run on a real Azure VM or App Service. The certificate path and the permission engine were exercised against a live tenant during `9295ef5`.

Reviewed by 15 adversarial rounds; residual non-blocking notes are in `docs/plans/azure-config-page-audit-report.md` and `docs/plans/azure-managed-identity-guided-setup.md` (local plan files).

### DIFF
--- a/docs/azure-setup.md
+++ b/docs/azure-setup.md
@@ -1,0 +1,70 @@
+# Azure setup
+
+Connapse reads Azure Blob Storage (including Data Lake Gen2 accounts) as its own Entra app
+identity, authenticated with a certificate it generates on its host, or as a managed identity.
+No account keys, SAS tokens, client secrets, or connection strings are ever entered or stored.
+It only reads: it never writes a blob, a role assignment, an ACL, or a tag.
+
+## Setup
+
+Admin → Providers → Azure. Two steps, each with an **Easy setup** guide and **Manual values**.
+
+1. **Access.** Copy the script, run it in [Azure Cloud Shell](https://shell.azure.com) (Bash),
+   paste back what it prints, Save. It uses the subscription Cloud Shell is signed in to, creates
+   two custom roles and the access app, and uploads only the public certificate:
+   - *Connapse Blob Data + Tags Reader* — read blobs and blob index tags, list containers.
+   - *Connapse RBAC Authorization Reader* — read role and deny assignments and storage-account
+     metadata at subscription scope.
+   Whoever runs it needs Owner or User Access Administrator on the subscription and permission
+   to register applications. The script is safe to re-run: it finds existing apps by name and
+   updates the roles in place.
+2. **Per-user permissions.** A second script registers the sign-in app people link their Entra
+   identity through, grants the access app the two Microsoft Graph permissions
+   (`User.Read.All`, `GroupMember.Read.All`) it looks people and groups up with, and attempts
+   admin consent. Consent needs a Global Administrator or Privileged Role Administrator; if the
+   person running the script is not one, the page shows a consent link to send them. When
+   consent succeeds the script also pre-consents the sign-in app for everyone, so nobody sees a
+   consent prompt when linking.
+
+Then Admin → Connections → New → Azure Blob Storage: choose the storage account and containers
+from the lists (or type them), Test connection, Save. People link their Entra identity under
+Integrations → Microsoft Entra ID; their search results are then limited to what that identity
+may read.
+
+## Troubleshooting
+
+**`AADSTS700027: The certificate ... is not registered on application`** — the certificate
+Connapse signs with is not the one uploaded to that app. Open the step's guide on the Azure
+provider page (it embeds the certificate Connapse actually holds), re-run the script, paste back,
+Save. Use **New certificate** only if you want to replace the key.
+
+**`AADSTS500113: No reply address is registered`** on the admin-consent link — the access app
+had no reply URL. Re-run the Per-user permissions script; it registers the Providers page as
+the reply URL and the link works.
+
+**`AADSTS65001` / "admin consent pending"** — the Graph permissions have not been approved.
+Send the consent link from the Per-user permissions card to a Global Administrator.
+
+**"This identity may not list storage accounts / containers"** in the connection form — the
+access identity lacks a listing role. Roles created before container/account listing was added
+are upgraded by re-running the Access script; otherwise assign *Connapse Blob Data + Tags Reader*
+on the account and *Connapse RBAC Authorization Reader* on the subscription. Typing the names
+still works without the lists.
+
+**Test connection fails with 403 / `AuthorizationPermissionMismatch`** — the access app has no
+blob-read role on that account or container. Use the "Role assignments granting exactly this
+access" snippet on the connection form, or assign *Connapse Blob Data + Tags Reader* at the
+account scope. Role assignments can take a few minutes to propagate.
+
+**Test connection fails with 404** — the container name is wrong, or the account is not the one
+you think (check the resource group in the account picker). Azure accounts have no default
+container; the test needs one that exists.
+
+**Search returns nothing for a linked user** — per-user filtering is on and the identity lookup
+failed closed. Check that admin consent was granted, that the subscription is set on the Access
+step (the RBAC resolver needs it), and that the user still exists and is enabled in Entra. For
+Data Lake Gen2 accounts, the user also needs Read on the file and Execute on every ancestor
+directory in the POSIX ACLs.
+
+**"Could not connect your Entra identity"** — the container log names the cause:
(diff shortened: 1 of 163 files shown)


Destrayon:
Codex adversarial review of the whole epic against `main` (`gpt-5.6-sol`; the plugin's sandbox refused to start processes today, so it ran via `codex exec` with the sandbox off and a read-only instruction — `git status` was clean afterwards). Verdict FAIL with seven findings. Verified against the code: four are defects, fixed in the commit above; three are design positions I'd like a decision on rather than a silent fix.

**Fixed**
1. An empty Azure configuration resolved to whatever managed identity the host had (a reset or lost row would quietly continue as the host). Nothing configured now fails closed; the host's identity is only ever the explicit flag. A certificate app beside a user-assigned identity throws instead of picking the certificate.
2. The blob endpoint override was independent of the account name: content could be read from one account, labelled and authorised as another, and Connapse's storage token sent to an arbitrary host. The override must now be the named account on an Azure cloud or a loopback emulator serving it — enforced on the form, connector, tester, and container picker.
4. Account casing: URIs and stored config are now lower case as Azure spells them, so a deny assignment cannot be missed by an uppercase name.
6. Disconnect now revokes every sign-in or parked confirmation the user started before it (a second tab could otherwise re-create the link — the same user's own identity, but still wrong).

**For decision (not changed)**
3. *Five-minute Graph/RBAC caches and a 60-second ancestor cache mean a revocation can take up to that long to bite.* This is the epic's documented design (`docs/superpowers/specs/2026-09-05-azure-blob-provider-design.md`: "revocation-propagation delay (= TTL)"), chosen for per-query cost. Options: keep and document the window in `docs/azure-setup.md`; shorten TTLs; or drop positive caching and pay two control-plane calls per query. I recommend keep + document, and measure under load as the spec already says.
5. *The host managed-identity detection reads token claims without validating the signature.* The claims only choose which principal the setup script names; they come from the local metadata endpoint, and anyone able to forge that answer already controls the host's credential. Validating the signature against the tenant's keys is doable as a follow-up (the Identity project has the JWT tooling) if you want defence in depth.
7. *No optimistic concurrency on the settings table* — two administrators saving the same form can overwrite each other. Product-wide and older than the epic; the Azure saves verify a candidate with Entra before storing, so a stale save is at worst a working-but-older credential. Worth its own issue (an ETag on `settings` + conflict on save) rather than this PR.

Non-blocking items taken: the token exchanger's failure message now carries the Entra error code and correlation id, not the response body; trailing whitespace fixed. Left: the in-memory sign-in stores are single-process by design (documented on the classes).

Destrayon:
Decisions on the three open review items, as recommended:

- **Cache windows (3):** kept. The propagation window is now documented in `docs/azure-setup.md` ("How quickly a permission change takes effect": at most five minutes for identity and role changes, sixty seconds for Gen2 directory traversals; tags and file ACLs read live).
- **Metadata-token signature (5):** follow-up #506 — validate the token against the tenant's keys and fail closed on missing claims; defence in depth, not a present exposure.
- **Settings concurrency (7):** follow-up #507 — an ETag/version on the `settings` row with a conflict on stale saves; product-wide, not specific to this epic.

